PR #1, opened 2015-04-25
Added SMTPS and ODMR

fowlslegs:
Added SMTPS and On-Demand Mail Relay (used by Fetchmail, xODMR). Both of these are in the Fedora 20 /etc/services.

### DIFF
--- a/qubesmanager/firewall.py
+++ b/qubesmanager/firewall.py
@@ -97,10 +97,10 @@ class NewFwRuleDlg (QDialog, ui_newfwruledlg.Ui_NewFwRuleDlg):
             ]
         displayed_services = [
                 '',
-                'http', 'https', 'ftp', 'ftps',
-                'smtp', 'pop3', 'pop3s', 'imap', 'imaps', 'nntp', 'nntps',
-                'ssh', 'telnet', 'telnets', 'ntp', 'snmp',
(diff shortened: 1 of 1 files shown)


marmarek:
Merged to my repo, thanks.
Will push here when we release new binaries update.

fowlslegs:
So in QM if you want smtps on F21 you need to enter in urd. If you select SMTPS you'll get an error. Problem is nobody knows smtps and URL Rendesvous Directoray (urd) share the same port (465) and you need to type in the latter to use it. I think we should keep smtps as a GUI option, but alias it to urd. I will see if I can make a pull request that would work this out nicely.

marmarek:
`/etc/services` define aliases, smtps is there. Maybe you could add
parsing of those aliases in qubes-manager/qubesmanager/firewall.py?

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab
A: Because it messes up the order in which people normally read text.
Q: Why is top-posting such a bad thing?

fowlslegs:
Okay, I do very much want to do this myself, but am only partway through Ch.4 of the Python book recommended by Qubes dev docs section. I think by the time I finish Ch.6 I should be able to write this. Beware, it will be my first Python code outside of the book exercises, but I've been told I'm a good coder once I get the hang of a language. I would just check it over carefully.

PR #2, opened 2015-05-24
Checking whether the "default NetVM" exists

nihilifer:
The "default NetVM" is usually the first created ProxyVM which is
set by qubes-core during its creation. [1] If there is no ProxyVM,
there is no "default NetVM". Therefore, creating an AppVM and
launching its settings dialog raised AttributeError, because
get_default_netvm method returned None.

This can be reproduced by installing QubesOS without creating VMs
by installer.

[1] https://github.com/QubesOS/qubes-core-admin/blob/master/core/qubes.py#L355

Fixes qubesos/qubes-issues#1008


marmarek:
Thanks. I've merged it to my repo, will push it here when releasing new package version.

PR #3, opened 2015-11-09
trimed exception popup message text quite a bit. Fixes QubesOS/qubes-…

bnvk:
…issues#1394

@marmarek how important / varried is the output of `exc_type.__name__` is it things other than `QubesException` which might be helpful to the user? I opted to trim that part out showing just the value of `exc_value` but could instead do something like

```
error = "%s: %s" % (exc_type.__name__, exc_value)
error = error.replace('QubesException: ', '')
```

### DIFF
--- a/qubesmanager/main.py
+++ b/qubesmanager/main.py
@@ -1985,8 +1985,10 @@ def handle_exception(exc_type, exc_value, exc_traceback):
 
     filename, line, dummy, dummy = traceback.extract_tb(exc_traceback).pop()
     filename = os.path.basename(filename)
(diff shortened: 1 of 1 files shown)


marmarek:
That's ok about 'QubesException: ' - this one will be removed, but any other exception will stay.

PS Please keep first line of commit message reasonable short (72 chars), and empty line before rest of the message. It's ok to put "Fixes ..." on a separate line.

bnvk:
> That's ok about 'QubesException: ' - this one will be removed, but any other exception will stay.

Ok. Pushed tweak to that!

marmarek:
Thanks!

PR #5, opened 2016-02-19
Clarify language in services tab

kbrn:
The services tab says that unlisted services will, "follow default VM's settings."  When I happened to see this, I was confused: which default VM?  The default NetVM?  The default TemplateVM?

After some investigating in this repo and in 000QubesVm.py, I've come to understand that they simply take on *default* values, not the values from some *default VM*.

If this is not correct, and this request should be rejected, the language should definitely still be clarified!


marmarek:
Thanks, will merge it somehow later this week.

PR #6, opened 2016-06-23
Implemented 'Restart' button

plushambush:
QubesOS/qubes-issues#1499


marmarek:
(for the next time) take a look at https://www.qubes-os.org/doc/code-signing/
We're going to enforce this rule soon (will be announced).

plushambush:
Will definitely do

PR #10, opened 2016-11-22
Require typing name of VM to remove

jpouellet:
IMO it is too easy to accidentally delete the wrong VM if you use lots of short-term (but longer than disposable) VMs and have built muscle memory of just clicking "Ok" in the deletion confirmation box.

This patch requires you to type the full name of the VM, inspired by repo deletion confirmation on GitHub.

The confirmation box looks like this:
confirm-delete

And if you type it wrong, you get this:
confirm-wrong

### DIFF
--- a/qubesmanager/main.py
+++ b/qubesmanager/main.py
@@ -1011,15 +1011,25 @@ class VmManagerWindow(Ui_VmManagerWindow, QMainWindow):
 
                 return
 
-        reply = QMessageBox.question(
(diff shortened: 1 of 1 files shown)


marmarek:
For reference: this is QubesOS/qubes-issues#1329

jpouellet:
Oh nice. Fixing issues without even knowing they exist!

Such is life when you have ~600 open...

PR #12, opened 2016-11-23
Implement searching for VMs in qubes-manager

jpouellet:
Example:
search-fedora

Some UX person (@bnvk?) might want to improve the placement or something, idk, but it's a HUGE usability improvement as-is for me, and likely anyone else who has 50+ VMs.

### DIFF
--- a/mainwindow.ui
+++ b/mainwindow.ui
@@ -59,6 +59,12 @@
     </property>
     <item row="0" column="0">
      <layout class="QHBoxLayout" name="horizontalLayout">
+      <property name="spacing">
+       <number>6</number>
+      </property>
+      <property name="leftMargin">
+       <number>6</number>
+      </property>
       <item>
        <widget class="QLabel" name="label">
         <property name="text">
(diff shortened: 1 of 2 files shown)


jpouellet:
Maybe it should be hidden by default and show up with a keyboard shortcut? (`Ctrl+F` or `/`)?

Regardless, IMO such changes should be done separately and not block this PR.

marmarek:
If it isn't hidden by default, should at least contain some label before ("Filter:" ?). I guess you've checked if QTableWidget have some built-in search option to re-use, right?

jpouellet:
> If it isn't hidden by default, should at least contain some label before ("Filter:" ?).

Good call. Will add that to this PR.

> I guess you've checked if QTableWidget have some built-in search option to re-use, right?

Well, there's [.findItems()](https://doc.qt.io/qt-4.8/qtablewidget.html#findItems), but I'd need to implement the same O(n) showing/hiding anyway, so this way seems cleaner to me.

jpouellet:
Now looks like this:
search-label

PR #14, opened 2016-11-26
Make inter-appviewer lock permissions consistent

jpouellet:
[Currently in gui-daemon/xside.c (only other holder of the lock):](https://github.com/QubesOS/qubes-gui-daemon/blob/95417c573d9b24269d50b3733164c3c9e390851c/gui-daemon/xside.c#L428-L435)

```
    g->inter_appviewer_lock_fd = open("/var/run/qubes/appviewer.lock",
        O_RDWR | O_CREAT, 0666);
    ...
    /* ignore possible errors */
    fchmod(g->inter_appviewer_lock_fd, 0666);
```

PR #15, opened 2016-11-27
Implement "Copy Dom0 clipboard" tray menu item

jpouellet:
Ctrl+Shift+C would be preferable, but is quite difficult to do safely,
and this method seems like a reasonable security/UX compromise.

Discussed in:
- https://groups.google.com/forum/#!topic/qubes-devel/4_x7Su21i9o
- https://github.com/QubesOS/qubes-issues/issues/2438#issuecomment-261704575

Fixes https://github.com/QubesOS/qubes-issues/issues/2450

### DIFF
--- a/qubesmanager/clipboard.py
+++ b/qubesmanager/clipboard.py
@@ -2,6 +2,7 @@
 #
 # The Qubes OS Project, http://www.qubes-os.org
 #
+# Copyright (C) 2016  Jean-Philippe Ouellet <jpo@vt.edu>
 # Copyright (C) 2012  Agnieszka Kostrzewa <agnieszka.kostrzewa@gmail.com>
 # Copyright (C) 2012  Marek Marczykowski <marmarek@mimuw.edu.pl>
 #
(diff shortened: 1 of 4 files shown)


jpouellet:
If we agree this is the way forward, I will document it in https://www.qubes-os.org/doc/copy-paste/

jpouellet:
Looks like this:
copy-to-dom0

jpouellet:
Note to people testing: copying out of dom0 in general is currently broken in general (including for "Copy to Qubes clipboard" button in the log viewer), see https://github.com/QubesOS/qubes-gui-daemon/pull/10 for one possible solution.

jpouellet:
This also (inadvertently) includes the change from https://github.com/QubesOS/qubes-manager/pull/14/files. I can separate it if you wish.

marmarek:
> This also (inadvertently) includes the change from https://github.com/QubesOS/qubes-manager/pull/14/files. I can separate it if you wish.

That's ok,

qubesmanager/clipboard.py
```diff
@@ -0,0 +1,56 @@
...
+        QMessageBox.warning (None, "Warning!", "Error while accessing Qubes clipboard!")
+        return
+
+    qubes_clipboard = open("/var/run/qubes/qubes-clipboard.bin", 'w')
+    qubes_clipboard.write(text)
+    qubes_clipboard.close()
+
+    qubes_clip_source = open("/var/run/qubes/qubes-clipboard.bin.source", 'w')
+    qubes_clip_source.write("dom0")
+    qubes_clip_source.close()
```
marmarek: This all should be in try/finally. Not releasing `appviewer.lock` will effectively break all VM GUI. I know it was "broken" before, but lets fix it while at it, especially when it's going to be used in one more place.

jpouellet: :+1: Will do

jpouellet:
^ rebased to master

marmarek:
> ^ rebased to master

Does it mean I should abstain from merging #14?

jpouellet:
> Does it mean I should abstain from merging #14?

If you intend to, then go ahead and I'll adjust this after.

jpouellet:
Done.

PR #18, opened 2016-12-18
Select search box contents on focus

jpouellet:


qubesmanager/main.py
```diff
@@ -465,6 +466,9 @@ def __init__(self, qvm_collection, blk_manager, parent=None):
 
         self.connect(self.searchbox, SIGNAL("textChanged(const QString&)"),
                      self.do_search)
+        self.searchbox.focusInEvent = types.MethodType(
+            lambda w, e: QTimer.singleShot(0, self.searchbox.selectAll),
+            self.searchbox.focusInEvent)
```
marmarek: How does it work? what about calling `super(...)`? Is really a timer required here?

jpouellet: I am not a Qt expert, but I was unable to get it working correctly with a simpler method.

I first tried subclassing QLineEdit and calling super() instead of binding like this, but then you must either deal with trying to instantiate non-default widget types in QtDesigner, or avoid doing so and grow VMManagerWindow.__init__ unnecessarily.

A timer is used because trying to select the text from in the context focusInEvent both does not work and causes unwanted side effects like the cursor not being visible. My knoledge of Qt is not deep enough to know why. Selecting after from the main event loop avoids these issues.

jpouellet: A timer is still required because the mousePressEvent gets handled after the focusInEvent, and the mousePressEvent deselects the text.

jpouellet:
An alternative way is this: https://github.com/QubesOS/qubes-manager/pull/20/files

marmarek:
Merged the alternative.

PR #19, opened 2016-12-18
Add some more keyboard shortcuts (Ctrl+_)

jpouellet:
As requested here: https://groups.google.com/d/topic/qubes-devel/YNqCYnxwgPI/discussion
**EDIT:** more discussion here https://groups.google.com/d/topic/qubes-devel/poVtf6X0-BM/discussion

shortcuts-1
shortcuts-2

Rationale:
- Ctrl+N for New, because it's nearly ubiquitous
- Ctrl+Backspace for delete vm, because that seems like a natural key to want to use to delete something
- Ctrl+C for Clone, because Clone and Copy start with C. Ctrl+Shift+N may also be a good candidate here, idk.
- Ctrl+Space for start & stop. We can use the same key because only one of those is enabled. Chosen because I use space to start & stop music and videos in nearly all players, just here it's a VM. Maybe just space without the Ctrl+ would be preferable?
- Ctrk+K for Kill, seems natural to me
- Ctrl+E for settings. Thinking is you are going to **e**dit the VM, and people are used to Ctrl+S being save
- Ctrl+U for Update, seems natural to me
- Ctrl+R for "Run command in VM" - already have a different way I do this, but seems to me like something people may want
- Ctrl+H for show/hide inactive VMs, seems natural, and for me very frequently used
- Ctrl+Shift+H for show/hide internal VMs, makes sense that the other show/hide should be like the first but with another modifier

Note that in https://github.com/QubesOS/qubes-manager/pull/13 I introduced the Qt standard Alt+key menu navigation, so all menus are already fully keyboard-usable, so these changes here are not intended to be for the most frequently used things, rather than comprehensive coverage.

I am open to suggestions for other keys. I am purposefully not assigning all keys here because once you assign a key and people start using it, it's really annoying to users to change it out from under them. IMO we should consider these assignments relatively immutable once they go into a release.

### DIFF
--- a/mainwindow.ui
+++ b/mainwindow.ui
@@ -416,6 +416,9 @@
    <property name="toolTip">
     <string>Create a new VM</string>
    </property>
+   <property name="shortcut">
+    <string>Ctrl+N</string>
+   </property>
   </action>
   <action name="action_removevm">
    <property name="enabled">
@@ -430,6 +433,9 @@
    </property>
    <property name="toolTip">
     <string>Remove an existing VM (must be stopped first)</string>
+   </property>
+   <property name="shortcut">
+    <string>Ctrl+Backspace</string>
    </property>
   </action>
   <action name="action_resumevm">
@@ -446,6 +452,9 @@
    <property name="toolTip">
     <string>Start/Resume selected VM</string>
    </property>
+   <property name="shortcut">
+    <string>Ctrl+Space</string>
+   </property>
   </action>
   <action name="action_pausevm">
    <property name="enabled">
@@ -461,6 +470,9 @@
    <property name="toolTip">
     <string>Pause selected VM</string>
    </property>
+   <property name="shortcut">
+    <string>Ctrl+P</string>
+   </property>
   </action>
   <action name="action_shutdownvm">
    <property name="enabled">
@@ -475,6 +487,9 @@
    </property>
    <property name="toolTip">
     <string>Shutdown selected VM</string>
+   </property>
+   <property name="shortcut">
+    <string>Ctrl+Space</string>
    </property>
   </action>
   <action name="action_restartvm">
@@ -521,6 +536,9 @@
    <property name="toolTip">
     <string>Update VM system</string>
    </property>
+   <property name="shortcut">
+    <string>Ctrl+U</string>
+   </property>
   </action>
   <action name="action_toggle_audio_input">
    <property name="enabled">
@@ -554,6 +572,9 @@
    </property>
    <property name="toolTip">
     <string>Show/Hide inactive VMs</string>
+   </property>
+   <property name="shortcut">
+    <string>Ctrl+H</string>
    </property>
   </action>
   <action name="action_editfwrules">
@@ -670,6 +691,9 @@
    <property name="toolTip">
     <string>VM Settings</string>
    </property>
+   <property name="shortcut">
+    <string>Ctrl+E</string>
+   </property>
   </action>
   <action name="action_restore">
    <property name="icon">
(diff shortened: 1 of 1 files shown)


qubenix:
utACK

I think most users are used to the media player shortcuts so no reason to reinvent that.

Is there no way to make Start/Pause VM on the same key? I think that would be more familiar. Then Shutdown VM could be moved to Ctrl+S.

jpouellet:
> I think most users are used to the media player shortcuts so no reason to reinvent that.

Are you proposing to use Space instead of Ctrl+Space?

> Is there no way to make Start/Pause VM on the same key? I think that would be more familiar. Then Shutdown VM could be moved to Ctrl+S.

It is **way** more likely that a user will want to shut down a VM than pause one.

Normal users will rarely (if ever) want to actually pause VMs at all. It does not consume less memory, AFAIK all it does is tell the scheduler not to run it. This is useful for ensuring you have a consistent state while dumping parts of it for debugging, and also as an anti-GUI-DoS mechanism. The former is exceedingly rare for normal users that it does not warrant being the default shortcut, and the latter is covered by the forthcoming Ctrl+Shift+P to pause all VMs.

marmarek:
What about using just "Del" for removing the VM? It would be much more intuitive, and since we've got better confirmation (thanks!), it should be safe.

Ctrl+Space for shutdown seems strange. But besides Ctrl+S I don't have any better idea. And Ctrl+S could be confused with Settings. 

Maybe better move the discussion to qubes-devel? Much more people are there than monitoring github comments...

jpouellet:
> What about using just "Del" for removing the VM? It would be much more intuitive, and since we've got better confirmation (thanks!), it should be safe.

Noted. I agree. My initial intention in prefixing everything with Ctrl was to avoid surprises when intending to input to search box but without focus there, and to preserve behavior of typing the beginning of a vm name to select the corresponding row in the table. Vim-style key bindings would be problematic here, but space & delete should be non-problematic.

> Ctrl+Space for shutdown seems strange. But besides Ctrl+S I don't have any better idea. And Ctrl+S could be confused with Settings.

The original thinking was space is intuitive given media player controls, and to prefix everything with ctrl. Although, think I prefer Space alone without modifier.

> Maybe better move the discussion to qubes-devel? Much more people are there than monitoring github comments...

Okay: https://groups.google.com/d/topic/qubes-devel/poVtf6X0-BM/discussion

woju:
Since main window just, well, went out the window, this should be summarily closed. Sorry for that.

@marmarek would you, please, since I don't have write permission to repo.

jpouellet:
Well, it may no longer exist in the new manager for 4.0, sure, but still exists in 3.2 which is still due to be supported for [quite a while](https://www.qubes-os.org/news/2016/09/02/4-0-minimum-requirements-3-2-extended-support/#extended-support-for-qubes-os-32).

I'm fine with abandoning my PR, but I'd like to better understand the reasoning because it sets a precedent for other PRs and issues as well. (for example https://github.com/QubesOS/qubes-issues/issues/2130, among many others of varying relevance)

Is the intention to focus exclusively on things which affect R4, and no longer expend energy maintaining R3.2 beyond bug fixes and security patches? Such a policy sounds fine to me, but if that's the case I think it should be made clear to users and other potential contributors to avoid wasted effort and mismatched expectations.

jpouellet:
To be clear, I'm not emotionally invested in this PR at all or trying to defend it, I'm just trying to understand the greater implications, if any.

marmarek:
Generally yes, stable release should receive only bug fixes. We do have flexible definition of "bug fixes", though.

jpouellet:
In that case I'm happy to pretend 4.0 is already released and 3.2 already in maintenance mode if it means effort can be focused there :)

jpouellet:
If that's the direction to take, there are many issues in https://github.com/QubesOS/qubes-issues/projects/3 with tag "enhancement" and specific to the old manager - these can perhaps be closed as WONTFIX as well.

PR #20, opened 2016-12-18
Select search box contents on focus (Attempt #2)

jpouellet:


qubesmanager/main.py
```diff
@@ -119,6 +119,23 @@ def process_IN_CREATE(self, event):
...
+
+    def focusInEvent(self, e):
+        super(SearchBox, self).focusInEvent(e)
+        self.selectAll()
+        self.focusing = True
+
+    def mousePressEvent(self, e):
+        super(SearchBox, self).mousePressEvent(e)
+        if self.focusing:
+            self.selectAll()
```
jpouellet: This one is required because the mousePressEvent otherwise deselects the text

marmarek:
While I see the downside of not having preview in QtDesigner, this one looks better. Maybe it's possible to set some minimal height of that layout, to at least have some hint how it looks like?

jpouellet:
> Maybe it's possible to set some minimal height of that layout, to at least have some hint how it looks like?

There is already the "Search:" label which accomplishes that:
qt-des-search
which I was actually going to remove in favor of [setPlaceholderText](https://doc.qt.io/qt-4.8/qlineedit.html#placeholderText-prop)("Search..."), but now perhaps not.

marmarek:
Ok, so this looks fine.

PR #23, opened 2017-01-26
Add extended information and Network diagram to Manager

unman:
QubesOS/qubes-issues#1431 
Functional (= not pretty) further information, and simple view of Qubes networking from About Menu


marmarek:
Not sure if this is a good place for such information. While for extended version info, it's a logical one, network information doesn't fit there. The whole about dialog is slightly too deep and to hard to find...
For the reference, this is how it looks:
qm-about-extra-info
qm-about-extra-version
qm-about-extra-network

marmarek:
@andrewdavidwong what do you think about it?

andrewdavidwong:
@marmarek: I agree with you. I do like the networking tree (nice feature!), but I think that it would be more appropriate as an item under the "System" tab.

PR #25, opened 2017-02-13
Proper fix for #2562

jpouellet:
https://github.com/QubesOS/qubes-issues/issues/2562

I'd meant to do this a while ago...

### DIFF
--- a/qubesmanager/main.py
+++ b/qubesmanager/main.py
@@ -719,7 +719,7 @@ class VmManagerWindow(Ui_VmManagerWindow, QMainWindow):
 
     @pyqtSlot(str)
     def do_search(self, search):
-        self.search = unicode(search)
+        self.search = str(search)
         self.showhide_vms()
         self.set_table_geom_size()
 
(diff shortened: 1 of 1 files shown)


marmarek:
> This regex is starting to get duplicated in too many places...
Maybe it belongs in core somewhere?

There is `vm.verify_name()` function, but to use that, you need to have VM instance already...

jpouellet:
Eh... IMO verify_name checks too much. The point here is to restrict the valid input chars, not so much that it would be a valid vm name. Consider for example someone with VM named "testing-dma-device" or such. Upon reaching "testing-d", typing the letter "m" would not produce a character in the text box because it would match verify_name's check that the name [cannot end in -dm](https://github.com/QubesOS/qubes-core-admin/blob/19e68bacf22cd965e819dfa4913740ecc14d1c40/core-modules/000QubesVm.py#L572).

jpouellet:
Perhaps more realistically, I sometimes search for "24" to match fedora-24, but verify_name [does not allow leading numbers](https://github.com/QubesOS/qubes-core-admin/blob/19e68bacf22cd965e819dfa4913740ecc14d1c40/core-modules/000QubesVm.py#L575), so I would need to type "-24" or such.

This is confusing and unexpected to users. Therefore, I think it's preferable as-is.

PR #31, opened 2017-04-03
Offset menu from mouse click

unman:
Resolves issue in QubesOS/qubes-issues#1911

### DIFF
--- a/qubesmanager/main.py
+++ b/qubesmanager/main.py
@@ -1758,7 +1758,7 @@ class VmManagerWindow(Ui_VmManagerWindow, QMainWindow):
             if self.blk_menu.isEmpty():
                 self.blk_menu.setEnabled(False)
 
-        self.context_menu.exec_(self.table.mapToGlobal(point))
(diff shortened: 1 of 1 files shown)


jpouellet:
Does this make it possible for click to register on different element than one under cursor if two ui elements are within 10px of each other?

unman:
On Mon, Apr 03, 2017 at 12:18:24AM -0700, Jean-Philippe Ouellet wrote:
> Does this make it possible for click to register on different element than one under cursor if two ui elements are within 10px of each other?
> 
No, it offsets the menu raise 10px from the point of click. The current
implementation raises the menu AT the clickpoint, resulting in mistaken
selection from menu.

woju:
Since main window is gone, this should be closed. Cc @marmarek.

jpouellet:
See comments here: https://github.com/QubesOS/qubes-manager/pull/19#issuecomment-314916101

unman:
@marmarek This is longstanding bug in Manager, which affects users on stable, I think it should be merged.

marmarek:
See #107

PR #34, opened 2017-05-28
Show size of current inter-VM clipboard in Ctrl-Shift-C notification

itinerarium:
QubesOS/qubes-issues#2825

Shows size in bytes if less than one KiB: 
> Copied **40 bytes** to the clipboard.

Shows size in bytes and units (up to GiB), rounded to one decimal place, if more than one KiB: 
> Copied **1904 bytes (1.9 KiB)** to the clipboard.

Are there any translation considerations for 'bytes'?

### DIFF
--- a/qubesmanager/clipboard.py
+++ b/qubesmanager/clipboard.py
@@ -24,6 +24,7 @@
 
 import os
 import fcntl
+from math import log
 
 from qubes.qubes import QubesException
 from PyQt4.QtGui import QApplication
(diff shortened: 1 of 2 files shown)


itinerarium:
Would it be possible to retry the policy/qubesos/code-signing check? The pub key has been added to GitHub. Thanks!

marmarek:
> Would it be possible to retry the policy/qubesos/code-signing check? The pub key has been added to GitHub. Thanks!

The script search the key from a keyserver. Have you uploaded it to any?

itinerarium:
@marmarek The key should now be available at http://keys.gnupg.net/pks/lookup?search=0xEF7D926F3F5ED819&fingerprint=on&hash=on&op=vindex. Thanks!

PR #36, opened 2017-07-12
Devel spring cleaning

woju:
Definitely a mess, almost certainly not bisectable, but it compiles, so let's ship it!

This includes new-api branch from @marmarta, with several minor fixes to that.

@marmarek, could you please pay special attention to firewall, since this one is actually not reworked much. Instead, it has a compatibility wrapper around, which is designed to barf an exception when it doesn't recognise something. This exception probably should be caught and should cause the firewall tab to be disabled somehow (this isn't done yet).


rpm_spec/qmgr.spec
```diff
@@ -11,11 +9,19 @@ Group:		Qubes
...
-Requires:	dbus
+Requires:	python3
+Requires:	python3-PyQt4
+Requires:	python3-inotify
+Requires:	qubes-core-dom0-linux >= 2.0.22
+Requires:	qubes-core-dom0 >= 3.0.18
 Requires:	qubes-artwork
-BuildRequires:	PyQt4-devel
+Requires:	pmount
+Requires:	cryptsetup
```
marmarek: Please drop those two - were needed for backup restore in R2...

qubesmanager/appmenu_select.py
```diff
@@ -20,120 +20,76 @@
...
+    def __init__(self, name, ident, parent=None):
         super(AppListWidgetItem, self).__init__(name, parent)
-        self.setToolTip(command)
-        self.filename = filename
+#       self.setToolTip(command)
+        self.ident = ident
 
+    @classmethod
+    def from_line(cls, line):
+        ident, name = line.strip().split(maxsplit=2)
```
woju: Yes, it should.

qubesmanager/backup.py
```diff
@@ -225,11 +225,11 @@ def shutdown_all_running_selected(self):
             self.blk_manager.check_if_serves_as_backend(vm)
 
         reply = QMessageBox.question(None, self.tr("VM Shutdown Confirmation"),
-             unicode(self.tr(
```
woju: No, I'm not sure. That was part of sed'ing for python3.

qubesmanager/settings.py
```diff
@@ -475,72 +400,48 @@ def check_mem_changes(self):
...
-        self.vcpus.setMaximum(QubesHost().no_cpus)
+#       self.vcpus.setMaximum(QubesHost().no_cpus)
         self.vcpus.setValue(int(self.vm.vcpus))
 
-        self.include_in_balancing.setEnabled(True)
-        self.include_in_balancing.setChecked(self.vm.services['meminfo-writer']==True)
-        if self.vm.type == "HVM":
-            self.include_in_balancing.setChecked(False)
-            self.include_in_balancing.setEnabled(False)
+        self.include_in_balancing.setEnabled(not self.vm.hvm)
```
woju: OK.

qubesmanager/settings.py
```diff
@@ -573,35 +474,37 @@ def __init_advanced_tab__(self):
                 self.drive_path.setText(drv_path)
             self.drive_domain.setCurrentIndex(self.drive_domain_idx)
 
-        if not hasattr(self.vm, "dispvm_netvm"):
```
woju: Oh, sure. Got dropped in one of the deletion sprints. Will bring it back.

rpm_spec/qmgr.spec
woju: Could you drop them yourself as part of redoing backups? Backup code is still there, so I think it should go kind of together.

qubesmanager/backup.py
marmarek: I'm pretty sure some of those dropped `unicode().format()` calls are reachable, please test.

rpm_spec/qmgr.spec
marmarek: Ok.

PR #41, opened 2017-09-13
Create new VM GUI

marmarta:
Reworked Create VM window

Redesigned for more logical flow; fixed problems with selecting
incompatible options.
fixes QubesOS/qubes-issues#1251

### DIFF
--- a/qubesmanager/create_new_vm.py
+++ b/qubesmanager/create_new_vm.py
@@ -26,6 +26,7 @@ import os
 import sys
 import threading
 import time
+import subprocess
 
 from PyQt4.QtCore import *
 from PyQt4.QtGui import *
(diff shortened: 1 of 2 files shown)


qubesmanager/create_new_vm.py
```diff
@@ -135,20 +151,72 @@ def accept(self):
...
+            try:
+                vm = app.add_new_vm(vmclass,
+                    name=name, label=label, template=template)
+                for k, v in properties.items():
+                    setattr(vm, k, v)
 
-        except Exception as ex:
-            thread_monitor.set_error_msg(str(ex))
+            except Exception as ex:
+                thread_monitor.set_error_msg(str(ex))
```
marmarek: Why not a single try/except over both cases?

PR #51, opened 2017-12-11
Backup gui restore

marmarta:
Restoration of backup-and-restore GUI. New API, some fixes to readability and usability, added 'compression' option.

fixes QubesOS/issues#3354


qubesmanager/backup.py
```diff
@@ -21,379 +20,299 @@
...
-class BackupVMsWindow(Ui_Backup, QWizard):
-
-    __pyqtSignals__ = ("backup_progress(int)",)
-
-    def __init__(self, app, qvm_collection, blk_manager, shutdown_vm_func, parent=None):
+    def __init__(self, app, qvm_collection, parent=None):
         super(BackupVMsWindow, self).__init__(parent)
 
         self.app = app
         self.qvm_collection = qvm_collection
```
marmarek: This may be misleading, as in many places `app` is used for `Qubes()`. IMO better use `qt_app` and `qubes_app` (or just `app`).

qubesmanager/backup.py
```diff
@@ -21,379 +20,299 @@
...
+            (lambda vm: vm.klass != 'TemplateVM' and vm.is_running()),
+            allow_internal=False,
+            allow_default=False,
+            allow_none=False
+        )
+
+        selected = self.load_settings()
+        self.__fill_vms_list__(selected)
 
     def load_settings(self):
```
marmarek: This function could use some docstring, explaining that it does (especially that it first try to load backup profile, then application specific settings).

qubesmanager/backup.py
```diff
@@ -21,379 +20,299 @@
...
+
+        if 'compression' in profile_data:
+            self.compress_checkbox.setChecked(profile_data['compression'])
+
+        if 'include' in profile_data:
+            return profile_data['include']
+
+        return None
+
+    def save_settings(self, use_temp):
```
marmarek: Docstring.

qubesmanager/backup_utils.py
```diff
@@ -20,90 +19,98 @@
...
     dialog.appvm_combobox.clear()
     dialog.appvm_combobox.addItem("dom0")
 
-    dialog.appvm_combobox.setCurrentIndex(0) #current selected is null ""
+    dialog.appvm_combobox.setCurrentIndex(0)  # current selected is null ""
 
-    for vm in dialog.qvm_collection.values():
-        if vm.is_appvm() and vm.internal:
+    for vm in dialog.qvm_collection.domains:
+        if vm.klass == 'AppVM' and vm.features.get('internal', False):
```
marmarek: I think 'internal' feature should be used regardless of VM class.

qubesmanager/backup_utils.py
```diff
@@ -20,90 +19,98 @@
...
-
-from .thread_monitor import *
+from . import utils
+import yaml
 
 path_re = re.compile(r"[a-zA-Z0-9/:.,_+=() -]*")
 path_max_len = 512
 
+
 def fill_appvms_list(dialog):
```
marmarek: Docstring, especially what is intended use of this function.

qubesmanager/backup_utils.py
```diff
@@ -20,90 +19,98 @@
...
-    dialog.appvm_combobox.setCurrentIndex(0) #current selected is null ""
+    dialog.appvm_combobox.setCurrentIndex(0)  # current selected is null ""
 
-    for vm in dialog.qvm_collection.values():
-        if vm.is_appvm() and vm.internal:
+    for vm in dialog.qvm_collection.domains:
+        if vm.klass == 'AppVM' and vm.features.get('internal', False):
             continue
-        if vm.is_template() and vm.installed_by_rpm:
+        if vm.klass == 'TemplateVM' and vm.installed_by_rpm:
```
marmarek: I'd drop this condition from here (ignore `installed_by_rpm`).

qubesmanager/backup_utils.py
```diff
@@ -20,90 +19,98 @@
...
-    if path_re.match(untrusted_path):
-        assert '../' not in untrusted_path
-        assert '\0' not in untrusted_path
-        return untrusted_path.strip()
-    else:
-        return None
-
-def select_path_button_clicked(dialog, select_file = False):
+
+def select_path_button_clicked(dialog, select_file=False, read_only=False):
```
marmarek: Docstring, especially parameters description.

qubesmanager/backup_utils.py
```diff
@@ -20,90 +19,98 @@
...
+    profile_data = {key: value for key, value in args.items()
+                    if key in acceptable_fields}
+
+    path = get_profile_path(use_temp)
+
+    with open(path, 'w') as profile_file:
+        yaml.safe_dump(profile_data, profile_file)
+
+
+def get_profile_name(use_temp):
```
marmarek: Logically this (and `get_profile_path`) should be before `load_backup_profile` and `write_backup_profile`.

qubesmanager/restore.py
```diff
@@ -126,119 +103,102 @@ def __fill_vms_list__(self):
...
-        if 'ignore-username-mismatch' in self.restore_options:
-            self.ignore_uname_mismatch.setChecked(self.restore_options['ignore-username-mismatch'])
-
-    def gather_output(self, s):
-        self.func_output.append(s)
+    def append_output(self, text):
+        self.commit_text_edit.append(text)
 
-    def restore_error_output(self, s):
+    def restore_error_output(self, text):
```
marmarek: Is this function used anywhere?

qubesmanager/restore.py
```diff
@@ -126,119 +103,102 @@ def __fill_vms_list__(self):
...
         self.error_detected.set()
-        self.feedback_queue.put((SIGNAL("restore_progress(QString)"),
-                                 u'<font color="red">{0}</font>'.format(s)))
+        self.append_output(u'<font color="red">{0}</font>'.format(text))
```
marmarek: Using `feedback_queue` was necessary here, because in Qt you should not touch GUI elements outside of main thread. So, feedback callback (called from another thread - see how you call `self.__do_restore__`) only queue the message, then post it to Qt from the main thread.

qubesmanager/restore.py
```diff
@@ -126,119 +103,102 @@ def __fill_vms_list__(self):
...
+    def __do_restore__(self, t_monitor):
         err_msg = []
-        self.qvm_collection.lock_db_for_writing()
         try:
-            backup.backup_restore_do(self.vms_to_restore,
-                                     self.qvm_collection,
-                                     print_callback=self.restore_output,
-                                     error_callback=self.restore_error_output,
-                                     progress_callback=self.update_progress_bar)
+            self.backup_restore.restore_do(self.vms_to_restore)
```
marmarek: New `BackupRestore.backup_do` use standard python logging instead of custom callbacks. You should register logger for `qubesadmin.backup` to catch those messages and display them. See here https://docs.python.org/3/library/logging.handlers.html#queuehandler
Then set it like this:
```python
handler = logging.handlers.QueueHandler(self.feedback_queue)
logger = logging.getLogger('qubesadmin.backup')
logger.addHandler(handler)
```
Handling elements in `self.feedback_queue` (in `current_page_changed`) may need to be adjusted for the format used by QueueHandler.

test-packages/qubesadmin/exc.py
```diff
@@ -1,4 +1,4 @@
-### mock qubesadmin.exc module
+# pylint: disable=unused-variable### mock qubesadmin.exc module
```
marmarek: Newline in between.

PR #55, opened 2018-01-12
 Added 'configure no strict reset for PCI devices' button 

marmarta:
The button is in 'devices' tab of VM settings; it allows the user to
select which devices should have 'no-strict-reset' enabled and tells
the user not to change this setting if there are no errors.
(Also, there are some minor formatting fixes, because pylint was unhappy)

fixes QubesOS/qubes-issues#3205

### DIFF
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -701,17 +701,16 @@ class VMSettingsWindow(ui_settingsdlg.Ui_SettingsDialog, QtGui.QDialog):
                             break
                     if current_assignment is None:
                         # it would be very weird if this happened
+                        msg.append(self.tr("Error re-assigning device ") +
+                                   ident)
                         break
+
                     self.vm.devices['pci'].detach(current_assignment)
 
-                    options = {}
-                    if ident in self.new_strict_reset_list:
-                        options['no-strict-reset'] = True
-                    new_assignment = devices.DeviceAssignment(
-                        self.vm.app.domains['dom0'],
-                        ident.replace(':', '_'),
-                        persistent=True, options=options)
-                    self.vm.devices['pci'].attach(new_assignment)
+                    current_assignment.options['no-strict-reset'] = \
+                        (ident in self.new_strict_reset_list)
+
+                    self.vm.devices['pci'].attach(current_assignment)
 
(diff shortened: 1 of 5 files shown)


qubesmanager/settings.py
```diff
@@ -664,27 +667,52 @@ def __init__(self, name, ident, parent=None):
...
+                elif (ident in self.current_strict_reset_list) != \
+                        (ident in self.new_strict_reset_list):
+                    current_assignment = None
+                    for assignment in self.vm.devices['pci'].assignments(
+                            persistent=True):
+                        if assignment.ident.replace("_", ":") == ident:
+                            current_assignment = assignment
+                            break
+                    if current_assignment is None:
+                        # it would be very weird if this happened
```
marmarek: Yes, it would. Better log it as an error. Currently it _silently_ abort applying PCI device changes...

qubesmanager/settings.py
```diff
@@ -664,27 +667,52 @@ def __init__(self, name, ident, parent=None):
...
+                        break
+                    self.vm.devices['pci'].detach(current_assignment)
+
+                    options = {}
```
marmarek: This should be `options = current_assignment.options`, otherwise you'll discard other options (if any).

qubesmanager/settings.py
```diff
@@ -664,27 +667,52 @@ def __init__(self, name, ident, parent=None):
...
+                    if ident in self.new_strict_reset_list:
+                        options['no-strict-reset'] = True
+                    new_assignment = devices.DeviceAssignment(
```
marmarek: You can re-use old assignment.

qubesmanager/settings.py
```diff
@@ -721,6 +749,19 @@ def devices_selection_changed(self):
                 self.dmm_warning_adv.hide()
                 self.dmm_warning_dev.hide()
 
+    def define_strict_reset_devices(self):
+        for assignment in self.vm.devices['pci'].assignments():
+            if 'no-strict-reset' in assignment.options and \
+                    assignment.options['no-strict-reset']:
```
marmarek: `assignment.options.get('no-strict-reset', False)`?

PR #56, opened 2018-01-16
Virt mode select fixes

marmarta:
Minor fixes to virt_mode select pull request

Changes:
- added hints about "which virt mode should I use"
- changed displayed virt modes to show modes in all caps
Based on pull request #44 

closes #44


qubesmanager/settings.py
```diff
@@ -619,12 +624,104 @@ def __apply_advanced_tab__(self):
         except qubesadmin.exc.QubesException as ex:
             msg.append(str(ex))
 
+        try:
+            if self.virt_mode.currentIndex() != self.virt_mode_idx:
+                self.vm.virt_mode = self.selected_virt_mode().lower()
+                # pylint: disable=attribute-defined-outside-init
+                self.anything_changed = True
```
marmarek: I think we've killed this attribute.

qubesmanager/settings.py
```diff
@@ -619,12 +624,104 @@ def __apply_advanced_tab__(self):
...
+        else:
+            devs_attached = bool(list(self.vm.devices['pci'].persistent()))
+
+        if devs_attached:
+            self.pvh_mode_hidden.show()
+        else:
+            choices.insert(0, 'PVH')
+            self.pvh_mode_hidden.hide()
+
+        if hasattr(self, 'virt_mode_list'):
```
marmarek: Then self don't have `virt_mode_list`? Anyway better always define attribute but set to `None` if not applicable.

qubesmanager/settings.py
```diff
@@ -619,12 +624,104 @@ def __apply_advanced_tab__(self):
...
+            old_mode = self.selected_virt_mode()
+            self.virt_mode.currentIndexChanged.disconnect()
+        else:
+            old_mode = None
+
+        self.virt_mode.clear()
+
+        # XXX: Hardcoded default value.
```
marmarek: `qubesadmin.QubesVM` now have `property_get_default()` method. When using it, add version dependency to spec file: `Requires: python3-qubesadmin >= 4.0.13`.
This is specifically relevant, because we've just changed the default to PVH...

qubesmanager/utils.py
```diff
@@ -33,7 +33,7 @@ def _filter_internal(vm):
 def prepare_choice(widget, holder, propname, choice, default,
         filter_function=None, *,
         icon_getter=None, allow_internal=None, allow_default=False,
-        allow_none=False):
+        allow_none=False, transform=None):
```
marmarek: Shouldn't transform be applied to `choices` and `default` too?

qubesmanager/utils.py
```diff
@@ -54,6 +54,8 @@ def prepare_choice(widget, holder, propname, choice, default,
 
     if propname is not None:
         oldvalue = getattr(holder, propname)
+        if transform is not None:
+            oldvalue = transform(oldvalue)
         is_default = holder.property_is_default(propname)
```
marmarek: (related to this place, not exactly the above line) IMO this would be a good place for holder.property_get_default() call.

PR #64, opened 2018-02-06
Fixed error in creating new VMs from GUI

marmarta:
Virt mode was forcibly set to HVM. Now, the virt mode is left as default,
unless a standalone VM with system installed from elsewhere is selected.

fixes QubesOS/qubes-issues#3515


qubesmanager/create_new_vm.py
```diff
@@ -120,8 +120,10 @@ def accept(self):
 
         properties = {}
         properties['provides_network'] = self.provides_network.isChecked()
-        properties['virt_mode'] = 'hvm'
         properties['netvm'] = self.netvm_list[self.netvm.currentIndex()]
+        if self.install_system.isChecked():
+            properties['virt_mode'] = 'hvm'
+            properties['kernel'] = 'none'
```
marmarek: None

PR #65, opened 2018-02-09
A bunch of Qube Manager fixes

marmarta:
- improved refreshing
- clear cache when it's needed
- hopefully FINALLY a fix for the shutdown bug

fixes QubesOS/qubes-issues#3501

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -256,7 +256,6 @@ class VmManagerWindow(ui_qubemanager.Ui_VmManagerWindow, QtGui.QMainWindow):
 
         self.vms_list = []
         self.vms_in_table = {}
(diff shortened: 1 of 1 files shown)


qubesmanager/qube_manager.py
```diff
@@ -395,6 +394,10 @@ def load_manager_settings(self):
     def get_vms_list(self):
         return [vm for vm in self.qubes_app.domains]
 
+    def update_single_row(self, vm):
+        if vm in self.qubes_app.domains:
```
marmarek: This looks suspicious. What if vm isn't in `self.qubes_app.domains`? Should be removed from the table?
And what if `vm.qid` is not in `self.vms_in_table`?

PR #73, opened 2018-02-21
Fix UnicodeErrors with translations and Unicode path's when restoring backup

donob4n:
Fix https://github.com/QubesOS/qubes-issues/issues/1859

Also needs a pull on qubes-core-admin

### DIFF
--- a/qubesmanager/backup_utils.py
+++ b/qubesmanager/backup_utils.py
@@ -72,7 +72,7 @@ def get_path_for_vm(vm, service_name):
         return None
 
 def select_path_button_clicked(dialog, select_file = False):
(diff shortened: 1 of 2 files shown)


donob4n:
Ouch, I just added my pgp key to git. What should I do?

marmarek:
Either [create signed tag](https://www.qubes-os.org/doc/code-signing/#using-pgp-with-git), or modify the top commit to be signed and force push: `git commit -S --amend && git push -f`

Also, upload the public key to some keyserver (`gpg --send-keys KEYID`).

donob4n:
Ok thanks.

PR #86, opened 2018-04-03
Srcrpm

fepitre:


fepitre:
@marmarek: how do you enable debug symbols when building the manager with python?

marmarek:
Similar issue: https://lists.opensuse.org/opensuse-buildservice/2015-11/msg00016.html

Generally it comes from lack of debug symbols - there is no such thing for python.

There are two options for python-only package:
1. `BuildArch: noarch`
2. `%global debug_package %{nil}`

The first one is the correct one, but we need to check if dnf will correctly handle upgrade. If not, lets go with the second one.

fepitre:
I will try to test an update with the noarch option.

fepitre:
I made a dummy update (version 4.0.18) and the update did well.

PR #90, opened 2018-05-03
Removed progress wait when updating template

donob4n:
Hi

I think this wait is pretty disruptive when you want to launch many updates at same time (e.g. debian-9, whonix-gw, whonix-ws). 

I tested without it some days and I had no problem. As a side effect the row is not marked as running until refreshing the manager, but this should be fixed when a DBUS event loop is enabled. Is anyone working on it?

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -919,20 +919,6 @@ class VmManagerWindow(ui_qubemanager.Ui_VmManagerWindow, QtGui.QMainWindow):
         thread.daemon = True
         thread.start()
 
-        progress = QtGui.QProgressDialog(
-                self.tr(
-                    "<b>{0}</b><br>Please wait for the updater to "
-                    "launch...").format(vm.name), "", 0, 0)
-        progress.setCancelButton(None)
-        progress.setModal(True)
-        progress.show()
-
-        while not t_monitor.is_finished():
-            self.qt_app.processEvents()
-            time.sleep(0.2)
-
-        progress.hide()
-
         if vm.qid != 0:
             if not t_monitor.success:
                 QtGui.QMessageBox.warning(


donob4n:
Hi

I already have a nearly ready dbus version for Qube Manager. I know the objective should be using Admin API events but until its problem has been solved (python async, KDE events loop, missing libs on Fedora repositories, ...) I think it's a better solution than current. It starts the window with all table ready, syncs with real state with few CPU usage and without the need of manual refresh ('Refresh' button is removed), and in any case it's a better place for migrate to other events/callbacks solution than current version.

Could I add a different pull request or better edit this one and commit the new changes?

Currently I'm trying to fix a bug with seems related to https://github.com/QubesOS/qubes-issues/issues/3856

marmarek:
> Could I add a different pull request or better edit this one and commit the new changes?

According to your description, it is unrelated to starting a template upgrade, so please create separate PR.

donob4n:
Thanks @marmarek , I will study that error cases.

About doing a separate PR, do I have to create a new branch for it on my repository?

marmarek:
> About doing a separate PR, do I have to create a new branch for it on my repository?

Yes.

donob4n:
> Maybe just removing progress dialog, but keeping all thread monitor stuff will work? That means processing Qt events only 5 times per second (sleep(0.2)), so may also not be such a good idea.

This is not ideal because processEvents() process events but user input is blocked. I am gonna try to send some kind of "error event" from the update thread to the main thread and let this last to show the dialog.

donob4n:
@marmarek 

I think the best solution is switching python threads to QThreads (maybe not needed but probably a more readable code) and use a 'ShowError' or 'HandleError' slot on main class connected to 'Error' SIGNALS on QThread classes.

So when a big task is needed just create the respective QThread, connect ERROR signal to ShowError and let's go.  All thread_monitor could be removed.

I would like to hear opinions before doing anything.

donob4n:
Since this needs more rewrite for doing it properly and I am currently finishing a dbus version I am gonna close for now and then propose a best solution.

PR #97, opened 2018-06-20
Log dialog separate

marmarta:
Log dialog viewer is now also available as a separate tool (introduced
to simplify displaying logs in the qui-domains tool)
Added entry point and main() to enable using log viewer as a separate tool.
Includes a minor fix for error that occurred with large log files due to changes
in Python 3.

### DIFF
--- a/qubesmanager/log_dialog.py
+++ b/qubesmanager/log_dialog.py
@@ -58,7 +58,7 @@ class LogDialog(ui_logdlg.Ui_LogDialog, QtGui.QDialog):
         if log.tell() > LOG_DISPLAY_SIZE:
             self.displayed_text = self.tr(
                 "(Showing only last %d bytes of file)\n") % LOG_DISPLAY_SIZE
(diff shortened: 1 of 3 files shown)


qubesmanager/log_dialog.py
```diff
@@ -56,7 +58,7 @@ def __init_log_text__(self):
         if log.tell() > LOG_DISPLAY_SIZE:
             self.displayed_text = self.tr(
                 "(Showing only last %d bytes of file)\n") % LOG_DISPLAY_SIZE
-            log.seek(-LOG_DISPLAY_SIZE, os.SEEK_END)
+            log.seek(log.tell()-LOG_DISPLAY_SIZE, os.SEEK_SET)
```
marmarek: What was wrong with the old version?

marmarta: Due to changes in file.seek() behavior in Python3, it didn't work at all - offset must be 0 with SEEK_END in case of text files. Reference: https://docs.python.org/3.2/library/io.html#io.TextIOBase.seek

PR #120, opened 2018-07-19
Handling VM dependencies in GUI tools

marmarta:
Added the following reactions to VM dependencies in GUI tools:
- Qube Manager will inform the user why they cannot delete a VM (which
properties of which VMs [or global] are using the VM)
- Settings window will try to intelligently rename VMs (change
properties to the new name, if possible, and inform the user what went
wrong if not)
- Settings window will inform the user why they cannot delete a VM
Also, renaming VM from Settings launched from Qube Manager will
refresh the VM list through a horrible hack, to be replaced by a neater
Admin API solution in some near future.

depends on QubesOS/qubes-core-admin-client@ca848ca

fixes QubesOS/qubes-issues#3992
fixes QubesOS/qubes-issues#3993
fixes QubesOS/qubes-issues#3892
fixes QubesOS/qubes-issues#3499

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -35,6 +35,7 @@ from pydbus import SessionBus
 
 from qubesadmin import Qubes
 from qubesadmin import exc
+from qubesadmin import utils
 
 from PyQt4 import QtGui  # pylint: disable=import-error
 from PyQt4 import QtCore  # pylint: disable=import-error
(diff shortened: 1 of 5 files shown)


qubesmanager/qube_manager.py
```diff
@@ -685,21 +691,25 @@ def action_removevm_triggered(self):
...
+                        manager_utils.format_dependencies_list(dependencies) + \
+                        "<br>"
+
+            info_dialog = QtGui.QMessageBox(self)
+            info_dialog.setWindowTitle(self.tr("Warning!"))
+            info_dialog.setText(
+                self.tr("This qube cannot be removed. It is used as:"
+                        " <br> {} <small>If you want to  remove this qube, "
+                        "you should remove or change settings of each qube "
+                        "or setting that uses it.</small>".format(list_text)))
```
marmarek: tr should get only format string, not already formatted message.

qubesmanager/settings.py
```diff
@@ -481,10 +482,35 @@ def _run_in_thread(self, func, *args):
...
+            failed_props = []
+
+            for (holder, prop) in dependencies:
+                try:
+                    if holder is None:
+                        setattr(self.vm.app, prop, new_vm)
+                    else:
+                        setattr(holder, prop, new_vm)
+                except qubesadmin.exc.QubesException as qex:
+                    failed_props += (holder, prop)
```
marmarek: += [ (...) ], otherwise you append this as two separate list elements

qubesmanager/settings.py
```diff
@@ -481,10 +482,35 @@ def _run_in_thread(self, func, *args):
...
+
+                QtGui.QMessageBox.warning(
+                    self,
+                    self.tr("Warning: rename partially unsuccessful"),
+                    self.tr("Some properties could not be changed to the new "
+                            "name. The system has now both {} and {} qubes. "
+                            "To resolve this, please check and change the "
+                            "following properties and remove the qube {} "
+                            "manually.<br> ".format(
+                                self.vm.name, name, self.vm.name) + list_text))
```
marmarek: Same about tr.

qubesmanager/settings.py
```diff
@@ -494,13 +520,31 @@ def _rename_vm(self, t_monitor, name):
...
+
+        if running_dependencies:
+            QtGui.QMessageBox.warning(
+                self,
+                self.tr("Qube cannot be renamed!"),
+                self.tr(
+                    "The following qubes using this qube as a template are "
+                    "running: <br> {}. <br> In order to rename this qube, you "
+                    "must first shut them down.".format(
+                        ", ".join(running_dependencies))))
```
marmarek: Same about tr.

qubesmanager/settings.py
```diff
@@ -516,6 +560,21 @@ def _remove_vm(self, t_monitor):
...
+
+        if dependencies:
+            list_text = utils.format_dependencies_list(dependencies)
+            QtGui.QMessageBox.warning(
+                self,
+                self.tr("Qube cannot be removed!"),
+                self.tr("This qube cannot be removed. It is used as:"
+                        " <br> {} <small>If you want to  remove this qube, "
+                        "you should remove or change settings of each qube "
+                        "or setting that uses it.</small>".format(list_text)))
```
marmarek: Same about tr.

RefinedSoftwareLLC:
With R4.0, security-testing. I was unable to delete a "-dvm" qube because its "Default DispVM" was set to itself. Since the delete command will also delete the offending qube (itself) this shouldn't prevent the deletion.

P.S. Error message said to look at system logs, but didn't say which qube (I assume Dom0), and it didn't say which log file (Dom0 has many files in the system log folder.)

PR #127, opened 2018-09-21
Fixed kernelopts in VM settings

marmarta:
For some reason, they were not being displayed at all. Now fixed.

fixes QubesOS/qubes-issues#4188


qubesmanager/settings.py
```diff
@@ -648,6 +648,7 @@ def __init_advanced_tab__(self):
                 None,
                 allow_default=True, allow_none=True)
             self.kernel.currentIndexChanged.connect(self.kernel_changed)
+            self.kernel_opts.setText(getattr(self.vm, 'kernelopts', '[]'))
```
marmarek: Why `[]` as default value?

marmarta: for consistency - the [] was default content when there were no kernel opts before.

PR #132, opened 2018-10-20
Switching to QThreads process

donob4n:
Hi,

There are some things to fix yet but I would like to comment and test travis.

The main change is using QThreads for all huge tasks. Threads are connected to a clear_threads() method which finds the finished thread on a list. if it has a msg set, opens a dialog with it, and if 'self.progress' is defined it hides it. I think it's simpler and more close to QT developers view.

I created a common_threads.py file (well, is the rewritten and renamed thread_monitor.py), for few threads that uses both qube_manager and settings. Do you find better to define them in one of this and delete the shared file?

I also improved ProgressDialog's and other minor things.


marmarek:
Yes, pylint complain about classes with only one public method - it may mean bad class design. In this case, it's because pylint doesn't see base QThread class (PyQt isn't installed in that environment), so it is a false positive. You can fix this by adding "pylint: disable=too-few-public-methods" comment at the beginning of such class.

donob4n:
Ah, thanks @marmarek 
I will continue this week.

donob4n:
I think this is ready.

What does ... ? https://github.com/QubesOS/qubes-manager/blob/77cf9bf1d23301a54f2849c0c5d8ce980b14fc0a/qubesmanager/restore.py#L181

I think it could be removed but since I am not sure about its purpose...

donob4n:
@marmarek Could you take a look on this? What do you think about ... ?
`old_sigchld_handler = signal.signal(signal.SIGCHLD, signal.SIG_DFL)`

marmarek:
> What does ... ?
> 
> [qubes-manager/qubesmanager/restore.py](https://github.com/QubesOS/qubes-manager/blob/77cf9bf1d23301a54f2849c0c5d8ce980b14fc0a/qubesmanager/restore.py#L181)
> 
> Line 181 in [77cf9bf](/QubesOS/qubes-manager/commit/77cf9bf1d23301a54f2849c0c5d8ce980b14fc0a)

git blame show this: https://github.com/QubesOS/qubes-manager/commit/1e7eaf37767bfe6a3c162a63e7b8dc22202dbeb0
That was a long time ago and since then the restore code was at least migrated to python3. But it still uses multiple processes and I don't see any obvious reason why this problem wouldn't apply anymore (if it applies back then). I'd leave it as it is, if it doesn't cause any specific problems.

donob4n:
@marmarek 
> > What does ... ?
> > [qubes-manager/qubesmanager/restore.py](https://github.com/QubesOS/qubes-manager/blob/77cf9bf1d23301a54f2849c0c5d8ce980b14fc0a/qubesmanager/restore.py#L181)
> > Line 181 in [77cf9bf](/QubesOS/qubes-manager/commit/77cf9bf1d23301a54f2849c0c5d8ce980b14fc0a)
> 
> git blame show this: [1e7eaf3](https://github.com/QubesOS/qubes-manager/commit/1e7eaf37767bfe6a3c162a63e7b8dc22202dbeb0)
> That was a long time ago and since then the restore code was at least migrated to python3. But it still uses multiple processes and I don't see any obvious reason why this problem wouldn't apply anymore (if it applies back then). I'd leave it as it is, if it doesn't cause any specific problems.

I just tested and I am pretty sure that currently it does nothing  because before calling it the handler is already SIG_DFL (both opening it from Qube Manager or directly), so it does not change anything.

@marmarta 
> Yeah, I was thinking about changing to model/view too - the current model is basically ported straight out of 3.2, and it's not exactly elegant. It would at least be more maintainable and I think at least not worse than it currently is - there's a bit less drawing/redrawing of the table than it used to be but it's still not very good.

Great, I would have to read something more about it but I hope to prepare a PR soon.

marmarek:
> I just tested and I am pretty sure that currently it does nothing because before calling it the handler is already SIG_DFL (both opening it from Qube Manager or directly), so it does not change anything.

Ok, in that case it can be removed.

qubesmanager/common_threads.py
```diff
@@ -0,0 +1,54 @@
...
+class RemoveVMThread(QtCore.QThread):
+    def __init__(self, vm):
+        QtCore.QThread.__init__(self)
+        self.vm = vm
+        self.msg = None
+
+    def run(self):
+        try:
+            del self.vm.app.domains[self.vm.name]
+        except exc.QubesException as ex:
```
marmarta: I wonder if - if a user tries hard enough or deletes a VM twice in a quick succession - they wont be able to get a KeyError here? I haven't managed to get it, but I wonder if it wouldn't be a good idea to except for that one, too.

qubesmanager/qube_manager.py
```diff
@@ -238,6 +243,56 @@ def check_if_vm_has_shutdown(self):
...
+            if self.vm.qid == 0:
+                subprocess.check_call(
+                    ["/usr/bin/qubes-dom0-update", "--clean", "--gui"])
+            else:
+                if not self.vm.is_running():
+                    self.vm.start()
+                self.vm.run_service("qubes.InstallUpdatesGUI",\
+                        user="root", wait=False)
+        except (ChildProcessError, exc.QubesException) as ex:
+            self.msg = ("Error on Qube update!", str(ex))
```
marmarta: I think in most other places 'qube' is not capitalized, I would not capitalize it here.

qubesmanager/qube_manager.py
```diff
@@ -432,21 +513,14 @@ def check_updates(self):
                     pass
 
     def on_domain_added(self, _submitter, _event, vm, **_kwargs):
+        domain = self.qubes_app.domains[vm]
```
marmarta: Have you tested this with a dispvm that fails to start? I think this can cause a KeyError or QubesException if the VM is deleted immediately after failing to start.

donob4n:
Thanks for review @marmarta , I will try to check it and fix tomorrow.

qubesmanager/qube_manager.py
donob4n: Yes, fixed.

qubesmanager/qube_manager.py
donob4n: There are other places with capitalized Qube, I will change them too.

donob4n: Oh, well maybe better I can wait for determine exactly when use one or other. e.g.
https://github.com/QubesOS/qubes-manager/blob/77cf9bf1d23301a54f2849c0c5d8ce980b14fc0a/qubesmanager/qube_manager.py#L690

qubesmanager/qube_manager.py
donob4n: I noticed my fix causes addition of empty rows if the code raises an exception after creating it. I have to improve it.

qubesmanager/backup.py
```diff
@@ -310,65 +323,56 @@ def current_page_changed(self, page_id): # pylint: disable=unused-argument
...
-                    self.turn_off_checkbox.isChecked():
+            if self.turn_off_checkbox.isChecked():
                 os.system('systemctl poweroff')
 
-        signal.signal(signal.SIGCHLD, old_sigchld_handler)
-
     def reject(self):
         if self.currentPage() is self.commit_page:
-            self.canceled = True
+            self.thread.terminate()
```
marmarek: Better call `admin.backup.Cancel` (the line below) and wait until it really is canceled - the `admin.backup.Execute` method should exit with exception, so there is no need to cancel that thread.

qubesmanager/common_threads.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/python2
+#
+# The Qubes OS Project, http://www.qubes-os.org
+#
+# Copyright (C) 2011  Marek Marczykowski <marmarek@mimuw.edu.pl>
```
marmarek: Shouldn't this be you?

qubesmanager/settings.py
```diff
@@ -150,6 +225,25 @@ def __init__(self, vm, qapp, init_page="basic", parent=None):
...
+                    self.progress = None
+
+                if thread.msg:
+                    (title, msg) = thread.msg
+                    QtGui.QMessageBox.warning(
+                        None,
+                        self.tr(title),
+                        self.tr(msg))
+
+                self.threads_list.remove(thread)
```
marmarek: Modifying the list while iterating over it is a bad idea. Iterate over its copy.

qubesmanager/qube_manager.py
```diff
@@ -410,12 +471,32 @@ def __init__(self, qt_app, qubes_app, dispatcher, parent=None):
...
+                    self.progress = None
+
+                if thread.msg:
+                    (title, msg) = thread.msg
+                    QtGui.QMessageBox.warning(
+                        None,
+                        self.tr(title),
+                        self.tr(msg))
+
+                self.threads_list.remove(thread)
```
marmarek: Modifying the list while iterating over it is a bad idea. Iterate over its copy.

qubesmanager/common_threads.py
donob4n: Well, this file was thread_monitor.py while I was working on it and when I deleted thread_monitor class I renamed it.

Really I think that it could be removed moving the code to qube_manager.py and import from there. What do you think?

marmarek: > Well, this file was thread_monitor.py while I was working on it and when I deleted thread_monitor class I renamed it.

There was nothing else in thread_monitor.py, besides ThreadMonitor class, so this is really a new file.

> Well, this file was thread_monitor.py while I was working on it and when I deleted thread_monitor class I renamed it.

This would create circular dependency (qube_manager.py already imports settings.py), so better avoid it. This isn't a long file, but I think it still makes sense to keep it separate (maybe more such classes will be added there in the future).

qubesmanager/settings.py
donob4n: Theorically we should have one call to 'clear_threads' for each finished thread so I added a return after removing and added a exception raise if something goes wrong.

PR #135, opened 2018-11-09
Template manager

marmarta:
A working template manager; at the moment it only provides means
to easily change templates of multiple VMs at once, but it should also give tools to install new templates in the future. It's also linked inside qube manager and the main system menu.

fixes QubesOS/qubes-issues#4085


marmarta:
It should be working now.

marmarek:
All of the above issues now are fixed.

There is one corner case though: if the qube is selected, then started, and then you choose "change all selected to", it does not work (no change to other qubes, does not reset to "(select template)"). And in the background you get this error:
```
  File "/usr/lib/python3.5/site-packages/qubesmanager/template_manager.py", line 182, in change_all_changed
    if row.checkbox.isChecked():
RuntimeError: wrapped C/C++ object of type QCheckBox has been deleted
```

Similarly, if the qube have new template selected, then is started and then you click "OK", you'll get:
```
Errors encountered on template change in the following qubes: 
vault: wrapped C/C++ object of type NewTemplateItem has been deleted.
```

PR #143, opened 2019-01-11
Fix to updating Qube Manager

marmarta:
Fixed label not updating and changed the update process to be more selective: now, the event
that fired the update is examined and only matching widgets are updated.

fixes QubesOS/qubes-issues#3675

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -143,6 +143,7 @@ class VmRowInTable:
         :return: None
         """
         try:
+            self.label_widget.update()
             self.info_widget.update_vm_state()
             self.template_widget.update()
             self.netvm_widget.update()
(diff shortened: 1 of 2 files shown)


qubesmanager/qube_manager.py
```diff
@@ -135,21 +135,31 @@ def __init__(self, vm, row_no, table):
...
+            if not event or event.endswith('netvm'):
+                self.netvm_widget.update()
+            if not event or event.endswith('internal'):
+                self.internal_widget.update()
+            if not event or event.endswith('ip'):
+                self.ip_widget.update()
+            if not event or event.endswith('include_in_backups'):
+                self.include_in_backups_widget.update()
+            if not event or event.endswith('backup_timestamp'):
+                self.last_backup_widget.update()
```
marmarek: Theoretically this will lead to extra updates if name of one property is a suffix of another. You should match even name including ":" separator.

marmarta: fixed; I don't know why github is not showing this as changed (maybe because I amended a commit).

PR #146, opened 2019-01-17
global_settings: Use MiB, not M as the unit

m-v-b:
This pull request has a dependency on the following one: https://github.com/QubesOS/qubes-core-admin/pull/249


m-v-b:
(It looks like there was a glitch with the verification of this commit's signature. My other pull request did not have this issue, despite having used the same key to sign both of the commits.)

PR #151, opened 2019-02-06
Added "set as dvm template" to VM settings

marmarta:
Added an 'Advanced' settings checkbox that allows the user to set a given
VM as template fo dispvms. It sets both the template_for_dispvms property
and the appmenus-dispvm feature - instead of delving into particulars
of how and which should be set, the user can just turn off the
whole setting if they want to make some changes in the vm and then turn
it on again.

fixes QubesOS/qubes-issues#4004

### DIFF
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -704,6 +704,11 @@ class VMSettingsWindow(ui_settingsdlg.Ui_SettingsDialog, QtGui.QDialog):
         self.seamless_on_button.clicked.connect(self.enable_seamless)
         self.seamless_off_button.clicked.connect(self.disable_seamless)
 
+        if hasattr(self.vm, "template_for_dispvms"):
+            self.dvm_template_checkbox.setChecked(self.vm.template_for_dispvms)
+        else:
+            self.dvm_template_checkbox.setVisible(False)
+
     def enable_seamless(self):
         self.vm.run_service_for_stdio("qubes.SetGuiMode", input=b'SEAMLESS')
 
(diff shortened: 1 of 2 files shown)


qubesmanager/settings.py
```diff
@@ -758,6 +763,16 @@ def __apply_advanced_tab__(self):
         except Exception as ex:  # pylint: disable=broad-except
             msg.append(str(ex))
 
+        if getattr(self.vm, "template_for_dispvms", False) != \
+                self.dvm_template_checkbox.isChecked():
+            try:
+                self.vm.template_for_dispvms = \
+                    self.dvm_template_checkbox.isChecked()
+                self.vm.features["appmenus-dispvm"] = \
+                    1 if self.dvm_template_checkbox.isChecked() else ""
```
marmarek: In case of disabling this setting, I'd remove `appmenus-dispvm` feature, instead of setting it to false.

BTW setting features to `bool` values is supported - they will be converted to appropriate values (`1`, empty) preserving true/false semantics.

ui/settingsdlg.ui
```diff
@@ -825,6 +825,16 @@ The qube must be running to disable seamless mode; this setting is not persisten
               </item>
              </layout>
             </item>
+            <item row="1" column="0" colspan="2">
+             <widget class="QCheckBox" name="dvm_template_checkbox">
+              <property name="toolTip">
+               <string/>
+              </property>
+              <property name="text">
+               <string>Allow starting Disposable VMs from this qube</string>
```
marmarek: This is yet another example why we need to settle on a solid definitions of VM vs Qube. I'll comment on #1015

PR #154, opened 2019-02-16
Added busy cursor to settings dialog

donob4n:
Hi,
It could be better but it would need pretty big rewrite.

fixes https://github.com/QubesOS/qubes-issues/issues/4744


marmarek:
Can you provide some timing data? According to @andrewdavidwong [guidelines](https://github.com/QubesOS/qubes-issues/issues/4744#issuecomment-464377415) progress bar is useful if the operation takes over 2s. I haven't managed to make it even over 1s. But it may be that my system is faster...

donob4n:
Yes I did some test and doing almost any possible modification I got between 3 and 4 seconds.

I will improve it with @andrewdavidwong suggestions.

donob4n:
Since it is not likely to spend too much time, the busy cursor seems enough notification so I removed the progress bar.

donob4n:
Don't merge this yet @marmarek I will add busy cursors to some dialogs which take few seconds for appearing (Global settings, Manage templates, Backup,...)

donob4n:
I think it's ready @marmarek , I hope it does not conflict with others PR.

PR #156, opened 2019-02-23
Fixed incorrect behavior after 'Apply' in Settings

marmarta:
Changes in some fields (to be precise, drop-downs) were discarded
in VM Settings when changed after an 'Apply' and before an 'OK'.

fixes QubesOS/qubes-issues#4833


qubesmanager/settings.py
```diff
@@ -1060,6 +1078,7 @@ def __remove_service__(self):
         del self.new_srv_dict[str(item.text())]
 
     def __apply_services_tab__(self):
+        # pylint
```
marmarek: Stray comment.

PR #167, opened 2019-03-29
Qube Manager widget update

marmarta:
- widgets were rewritten to have less duplicate code and make adding a new widget more straightforward
- a new widget that shows whether a VM is a DVM template was added

fixes QubesOS/qubes-issues#4912

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -126,7 +126,7 @@ class VmRowInTable:
 
         self.include_in_backups_widget = table_widgets.VMPropertyItem(
             vm, "include_in_backups",
-            empty_function=(lambda x: True if x is None or not x else False))
+            empty_function=(lambda x: not bool(x)))
         table.setItem(row_no, VmManagerWindow.columns_indices[
             'Include in backups'], self.include_in_backups_widget)
 
@@ -134,6 +134,11 @@ class VmRowInTable:
             vm, "backup_timestamp")
         table.setItem(row_no, VmManagerWindow.columns_indices[
             'Last backup'], self.last_backup_widget)
+
+        self.dvm_template_widget = table_widgets.VMPropertyItem(
+            vm, "template_for_dispvms", empty_function=(lambda x: not x))
+        table.setItem(row_no, VmManagerWindow.columns_indices['DVM Template'],
+                      self.dvm_template_widget)
 
         self.table = table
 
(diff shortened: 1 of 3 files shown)


qubesmanager/table_widgets.py
```diff
@@ -281,26 +264,33 @@ def update_vm_state(self):
         self.upd_info.update_outdated()
 
 
-class VmTemplateItem(QtGui.QTableWidgetItem):
-    def __init__(self, vm):
-        super(VmTemplateItem, self).__init__()
+class VMPropertyItem(QtGui.QTableWidgetItem):
+    def __init__(self, vm, property_name, empty_function=(lambda x: False),
+                 check_default=False):
```
marmarek: This could use some docstring describing arguments. What `empty_function` should do (and with what arguments it's called)? Same for `check_default`?

qubesmanager/table_widgets.py
```diff
@@ -281,26 +264,33 @@ def update_vm_state(self):
...
-        if getattr(self.vm, 'template', None) is not None:
-            self.setText(self.vm.template.name)
+        val = getattr(self.vm, self.property_name, None)
+        if self.empty_function(val):
+            self.setText("")
+        elif val is None:
+            self.setText("n/a")
+        elif self.check_default and \
+                self.vm.property_is_default(self.property_name):
+            self.setText('default (' + str(val) + ')')
```
marmarek: This, on bool property will be inconsistent with check_default=False. One with show `default (True)` and the other `Yes`.

qubesmanager/qube_manager.py
```diff
@@ -120,19 +120,26 @@ def __init__(self, vm, row_no, table):
...
-        self.last_backup_widget = table_widgets.VmLastBackupItem(vm)
+        self.last_backup_widget = table_widgets.VmLastBackupItem(
+            vm, "backup_timestamp")
         table.setItem(row_no, VmManagerWindow.columns_indices[
             'Last backup'], self.last_backup_widget)
 
+        self.dvm_template_widget = table_widgets.VMPropertyItem(
+            vm, "template_for_dispvms", empty_function=(lambda x: not x))
+        table.setItem(row_no, VmManagerWindow.columns_indices['DVM Template'],
+                      self.dvm_template_widget)
```
marmarek: I think QubesOS/qubes-issues#4912 is really about (badly named) `default_dispvm` property. Which should really be called `default_dvm_template` or such...

PR #168, opened 2019-04-02
Fixed device handling in VM settings widget

marmarta:
PCI devices are now handled as DeviceInfo objects, not (string,name) tuples;
furthermore, a device that no longer exists in the system can be
correctly removed via VM settings window (it is displayed as Unknown
device).

fixes QubesOS/qubes-issues#2789

### DIFF
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -926,27 +926,29 @@ class VMSettingsWindow(ui_settingsdlg.Ui_SettingsDialog, QtGui.QDialog):
         self.dev_list.add_all_button.setVisible(False)
         self.devices_layout.addWidget(self.dev_list)
 
-        devs = []
-        lspci = subprocess.check_output(['/usr/sbin/lspci']).decode()
-        for dev in lspci.splitlines():
-            devs.append((dev.rstrip(), dev.split(' ')[0]))
+        dom0_devs = list(self.vm.app.domains['dom0'].devices['pci'].available())
+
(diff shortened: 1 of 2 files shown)


qubesmanager/settings.py
```diff
@@ -926,27 +926,31 @@ def __init_devices_tab__(self):
...
 
-        devs = []
-        lspci = subprocess.check_output(['/usr/sbin/lspci']).decode()
-        for dev in lspci.splitlines():
-            devs.append((dev.rstrip(), dev.split(' ')[0]))
+        dom0_devs = [dev for dev in
+                     self.vm.app.domains['dom0'].devices['pci'].available()]
+
+        attached_devs = [dev for dev in
+                         self.vm.devices['pci'].persistent()]
```
marmarek: Those two could be simplified using `list()`, instead of list comprehension.

qubesmanager/settings.py
```diff
@@ -970,44 +974,43 @@ def __apply_devices_tab__(self):
         msg = []
 
         try:
-            old = [ass.ident.replace('_', ':')
-                   for ass in self.vm.devices['pci'].persistent()]
+            old_devs = [dev for dev in self.vm.devices['pci'].persistent()]
```
marmarek: And here too.

qubesmanager/settings.py
marmarta: why do you hate list comprehension????

PR #172, opened 2019-04-08
Add UI for managing Qubes update repositories

strugee:
Screenshot:

Screenshot_2019-04-08_03-24-50

Menu options provided:

* Stable updates
* Testing updates (security only)
* Testing updates
* Unstable updates
---
* ITL template updates
* ITL template updates (testing)
---
* (Community templates disabled)
* Community template updates
* Community template updates (testing)

I'll fix the TODO in the PR with the qrexec services and then delete the hack in this code. But since the rest of this is complete I figured I'd open a PR and get the ball rolling on code review.

Depends on QubesOS/qubes-core-admin-linux#48

Fixes QubesOS/qubes-issues#4550

### DIFF
--- a/qubesmanager/global_settings.py
+++ b/qubesmanager/global_settings.py
@@ -37,6 +37,51 @@ from configparser import ConfigParser
 
 qmemman_config_path = '/etc/qubes/qmemman.conf'
 
+def _run_qrexec_repo(service, arg=''):
+    # Fake up a "qrexec call" to dom0 because dom0 can't qrexec to itself yet
+    cmd = '/etc/qubes-rpc/' + service
+    p = subprocess.run(
+        ['sudo', cmd, arg],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
+    assert not p.stderr
+    assert p.returncode == 0
+    return p.stdout.decode('utf-8')
+
+def _manage_repos(repolist, action):
+    for i in repolist:
+        assert _run_qrexec_repo('qubes.repos.' + action, i) == 'ok\n'
+
+def _handle_dom0_updates_combobox(idx):
+    idx += 1
+    repolist = ['qubes-dom0-current', 'qubes-dom0-security-testing',
+         'qubes-dom0-current-testing', 'qubes-dom0-unstable']
+    enable = repolist[:idx]
+    disable = repolist[idx:]
+    _manage_repos(enable, 'Enable')
+    _manage_repos(disable, 'Disable')
+
+# pylint: disable=invalid-name
+def _handle_itl_tmpl_updates_combobox(idx):
+    idx += 1
+    repolist = ['qubes-templates-itl', 'qubes-templates-itl-testing']
+    enable = repolist[:idx]
+    disable = repolist[idx:]
+    _manage_repos(enable, 'Enable')
+    _manage_repos(disable, 'Disable')
+
+# pylint: disable=invalid-name
+def _handle_comm_tmpl_updates_combobox(idx):
+    # We don't increment idx by 1 because this is the only combobox that
+    # has an explicit "disable this repository entirely" option
+    repolist = ['qubes-templates-community',
+                'qubes-templates-community-testing']
+    enable = repolist[:idx]
+    disable = repolist[idx:]
+    _manage_repos(enable, 'Enable')
+    _manage_repos(disable, 'Disable')
+
 # pylint: disable=too-many-instance-attributes
 class GlobalSettingsWindow(ui_globalsettingsdlg.Ui_GlobalSettings,
                            QtGui.QDialog):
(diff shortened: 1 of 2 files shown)


strugee:
TODO is removed since I fixed the underlying bug in the other PR.

strugee:
Happy to squash the remaining PyLint warnings... but it seems unnecessary? The way it's currently written seems pretty clear and it's more concise.

marmarek:
While too long function name indeed is pylint being picky (you can add `#pylint: disable=invalid-name` just before it), I'd prefer to fix too short variable names to something more descriptive.

strugee:
@marmarek done! I squashed the changes into the existing PyLint commit.

FWIW since I forgot to mention this last night, I _have_ successfully tested these changes (including those I just pushed) decently thoroughly in dom0.

strugee:
@marmarek CI is passing.

strugee:
@marmarek ping for review, if you get a chance. No rush.

marmarek:
I haven't forgot, but first I want to get R4.1 in testable state. It's almost there: https://openqa.qubes-os.org/tests/overview?distri=qubesos&version=4.1&build=4.1-20190428&groupid=1 (see that one green dot!)

qubesmanager/global_settings.py
```diff
@@ -36,6 +37,52 @@
...
+        assert result == 'ok\n'
+
+def _handle_dom0_updates_combobox(idx):
+    idx += 1
+    repolist = ['qubes-dom0-current', 'qubes-dom0-security-testing',
+         'qubes-dom0-current-testing', 'qubes-dom0-unstable']
+    enable = repolist[:idx]
+    disable = repolist[idx:]
+    _manage_repos(enable, 'Enable')
+    _manage_repos(disable, 'Disable')
```
marmarek: Error in all those functions should result in readable error message. Something like "failed to enable/disable repository x" (which may be a text of raised exception inside _manage_repos). But it should be shown to the user in readable form, not generic "Huston, we have a problem...". Generally that handler is last resort, meaning some error handling is broken. See `VMSettingsWindow.save_changes` in `settings.py`. Other settings in this file lack proper error handling, so don't use them as an example...

qubesmanager/global_settings.py
```diff
@@ -251,6 +297,50 @@ def __init_updates__(self):
...
+
+        self.dom0_updates_repo.currentIndexChanged.connect(
+            _handle_dom0_updates_combobox
+        )
+        self.itl_tmpl_updates_repo.currentIndexChanged.connect(
+            _handle_itl_tmpl_updates_combobox
+        )
+        self.comm_tmpl_updates_repo.currentIndexChanged.connect(
+            _handle_comm_tmpl_updates_combobox
+        )
```
marmarek: All other settings here and in VM settings are applied on OK/Apply click. This shouldn't be an exception.

qubesmanager/global_settings.py
```diff
@@ -36,6 +37,52 @@
...
+def _run_qrexec_repo(service, arg=''):
+    # Fake up a "qrexec call" to dom0 because dom0 can't qrexec to itself yet
+    cmd = '/etc/qubes-rpc/' + service
+    p = subprocess.run(
+        ['sudo', cmd, arg],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
+    assert not p.stderr
+    assert p.returncode == 0
```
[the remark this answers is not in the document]
strugee: Ah yeah, I used `assert` because I assumed that that _would_ never happen, at least not if dom0 wasn't corrupted or misconfigured in some way... but I guess that qrexec policy could simply be set to denied by the administrator, right? Is that the impetus here?

qubesmanager/global_settings.py
strugee: Nice catch; fixed.

strugee:
@marmarek thanks very much for the helpful review, and sorry it took me a while to get around to making the changes - everything should be addressed. I also rebased while I was at it. The new changes have been manually tested so this should be ready to go in as soon as CI passes :thumbsup:

qubesmanager/global_settings.py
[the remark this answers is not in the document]
marmarek: Yes, policy is one possibility. But some configuration issue (missing repository definition for example) could happen too.

strugee:
OK, hopefully Travis is fixed (PyLint locally is complaining about short variable names, but it's in code I didn't touch recently... so we'll have to see what Travis says). I also found and fixed some bugs with the error handling, which is now fully tested. Here's what it looks like:

screenshot of a warning window in dom0 with the title "ERROR!" and the text "Error managing repository settings: qrexec call exited with non-zero return code; returncode: 1"

Also, before I forget, I want to point out that the program now unconditionally sets repository settings, even if they haven't changed. It was easier to write that way although if it's a big problem I can make it check beforehand.

marmarek:
> the program now unconditionally sets repository settings, even if they haven't changed.

No, that will be problematic, may unintentionally change setting that wasn't touched. And also will fail in case of strict qrexec policy (for example no repository change allowed, user wanted to change just default netvm).

strugee:
@marmarek fixed!

strugee:
Side note: PyQt really messes up debugging under Pdb; I used the method mentioned in [this Stack Overflow answer](https://stackoverflow.com/a/1745965/1198896) to fix it. According to the mailing list post linked in that answer it's always safe to call that method, so I wonder if we should just call it at application startup to improve developer experience. (I haven't really looked closely into what this method does, yet.)

Also, as a note to myself, I don't know if the functions that are now inside a class should be using `_` or `__` and I really should figure that out.

strugee:
Thanks @marmarek!! Super pumped to see this in a release :-)

PR #173, opened 2019-04-08
Made terminology in Create New VM GUI clearer

marmarta:
Changed the terms used in Create New VM Gui to the following:
-Qube based on a template (AppVM)
-Standalone qube
-Standalone qube copied from a template

references QubesOS/qubes-issues#4723


marmarek:
See https://github.com/QubesOS/qubes-issues/issues/4723#issuecomment-481058925

PR #175, opened 2019-04-30
Tests for manager

marmarta:
Extensive tests for Global Settings and Qube Manager. Can be normally run through python3 test_name.py.


marmarek:
While some test fails, generally it works, dosn't SEGV anymore :) 
https://openqa.qubes-os.org/tests/2404

PR #176, opened 2019-04-30
Assorted fixes

marmarta:
Fixes for a bunch of bugs discovered during test writing:
- incorrect check dom0 updates handling in Global Settings
- some unnecessary property setting in Global Settings
- no warning on attempts to clone a VM to an existing name in Qube Manager
- bug with hiding last visible column in Qube Manager
- protected Qube Manager from blowing up due to malformed settings file
- fixed logs menu contents in Qube Manager

### DIFF
--- a/qubesmanager/global_settings.py
+++ b/qubesmanager/global_settings.py
@@ -239,8 +239,8 @@ class GlobalSettingsWindow(ui_globalsettingsdlg.Ui_GlobalSettings,
         self.dom0_updates_file_path = '/var/lib/qubes/updates/disable-updates'
 
         try:
-            self.updates_dom0_val = self.qvm_collection.domains[
-                'dom0'].features['service.qubes-update-check']
+            self.updates_dom0_val = bool(self.qvm_collection.domains[
+                'dom0'].features['service.qubes-update-check'])
         except KeyError:
             self.updates_dom0_val =\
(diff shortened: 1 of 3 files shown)


qubesmanager/qube_manager.py
```diff
@@ -852,6 +863,19 @@ def action_clonevm_triggered(self):
         if not ok or clone_name == "":
             return
 
+        name_in_use = False
+        # check if VM name is already in use:
+        for domain in self.qubes_app.domains:
+            if domain.name == clone_name:
+                name_in_use = True
```
marmarek: Better check for existence in collection, not iterate over all domains: `clone_name in self.qubes_app.domains`

qubesmanager/global_settings.py
```diff
@@ -245,7 +245,7 @@ def __init_updates__(self):
             self.updates_dom0_val =\
                 not os.path.isfile(self.dom0_updates_file_path)
 
-        self.updates_dom0.setChecked(self.updates_dom0_val)
+        self.updates_dom0.setChecked(bool(self.updates_dom0_val))
```
marmarek: This should be converted to `bool` earlier. Otherwise check in `__apply_updates__` will fail (bool != str).

marmarek: BTW you can add issue reference: QubesOS/qubes-issues#4988

PR #178, opened 2019-05-22
Update Qt version used in qubesmanager to Qt5

marmarta:
Fixed dependencies, places where obsolete functions stopped working,
code fragments that started throwing warnings and an .ui file that
stopped being readable after the update.

### DIFF
--- a/qubesmanager/firewall.py
+++ b/qubesmanager/firewall.py
@@ -21,7 +21,7 @@
 import datetime
 import re
 
-from PyQt4 import QtCore, QtGui  # pylint: disable=import-error
+from PyQt5 import QtCore, QtGui, QtWidgets  # pylint: disable=import-error
 import qubesadmin.firewall
 
 from . import ui_newfwruledlg  # pylint: disable=no-name-in-module
@@ -29,6 +29,7 @@ from . import ui_newfwruledlg  # pylint: disable=no-name-in-module
 
 class FirewallModifiedOutsideError(ValueError):
     pass
+
 
 class QIPAddressValidator(QtGui.QValidator):
     # pylint: disable=too-few-public-methods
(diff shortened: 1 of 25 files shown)


qubesmanager/firewall.py
```diff
@@ -84,19 +86,20 @@ def __init__(self, parent=None):
...
-                msg = QtGui.QMessageBox()
-                msg.warning(self, self.tr("Firewall rule"),
+                msg = QtWidgets.QMessageBox()
+                msg.warning(
+                    self, self.tr("Firewall rule"),
                     self.tr("You need to fill service "
                             "name/port for TCP/UDP rule"))
                 return
-        QtGui.QDialog.accept(self)
+        QtWidgets.QDialog.accept(self)
```
marmarek: I think this should be `super().accept()`

qubesmanager/log_dialog.py
```diff
@@ -45,13 +44,15 @@ def __init__(self, app, log_path, parent=None):
...
-                     self.copy_to_clipboard_triggered)
+        self.copy_to_qubes_clipboard.clicked.connect(
+            self.copy_to_clipboard_triggered)
+        self.copy_to_qubes_clipboard.clicked.connect(
+            self.copy_to_clipboard_triggered)
 
         self.__init_log_text__()
 
     def __init_log_text__(self):
+        print(self.log_path)
```
marmarek: Leftover debug.

qubesmanager/log_dialog.py
```diff
@@ -62,16 +63,18 @@ def __init_log_text__(self):
         else:
             log.seek(0, os.SEEK_SET)
         self.displayed_text += log.read()
+        print(self.displayed_text)
```
marmarek: Leftover debug.

qubesmanager/restore.py
```diff
@@ -35,8 +34,7 @@
 from . import multiselectwidget
 from . import backup_utils
 
-from multiprocessing import Queue
-from multiprocessing.queues import Empty
+import queue
```
marmarek: `queue.Queue` is different from `multiprocessing.Queue`. According to documentation, the former works with threads, but not necessary with processes. Any specific reason for this change?

PR #182, opened 2019-06-05
Added 'provides network' checkbox to VM settings

marmarta:
A simple checkbox in the 'Advanced' tab. Does not allow disabling
'provides_network' if the VM is in use as NetVM; instead, a tooltip
informs the user which VMs use it.

fixes QubesOS/qubes-issues#5079

### DIFF
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -745,6 +745,17 @@ class VMSettingsWindow(ui_settingsdlg.Ui_SettingsDialog, QtGui.QDialog):
         else:
             self.dvm_template_checkbox.setVisible(False)
 
+        self.provides_network_checkbox.setChecked(
+            getattr(self.vm, 'provides_network', False))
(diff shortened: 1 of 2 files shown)


qubesmanager/settings.py
```diff
@@ -745,6 +745,20 @@ def __init_advanced_tab__(self):
         else:
             self.dvm_template_checkbox.setVisible(False)
 
+        self.provides_network_checkbox.setChecked(
+            getattr(self.vm, 'provides_network', False))
+        if self.provides_network_checkbox.isChecked():
+            domains_using = []
+            for domain in self.vm.app.domains:
```
marmarek: There is `vm.connected_vms` you can iterate on, or simply get as a list (call `list()` on it, it is a generator).

marmarta: pushed a fix (also, I express strong disapproval about the name; it is not clear if it refers to all dependent VMs or just those that use the VM as NetVM)

PR #186, opened 2019-07-07
Renamed DispVM and "Is DispVM template" for more readability

marmarta:
Replaced instances of "Default DispVM" and "Is DVM Template" with
more readable "Default Disposable VM Template" and "Disposable VM Template"
respectively. Added tooltips.

fixes QubesOS/qubes-issues#4935

### DIFF
--- a/ui/qubemanager.ui
+++ b/ui/qubemanager.ui
@@ -238,14 +238,14 @@ in backups</string>
       </column>
       <column>
        <property name="text">
-        <string>Default
-DisposableVM</string>
+        <string>Default DisposableVM
+Template</string>
        </property>
       </column>
       <column>
        <property name="text">
-        <string>Is template for 
-DisposableVMs</string>
+        <string>DisposableVM
+Template</string>
        </property>
       </column>
      </widget>
@@ -258,7 +258,7 @@ DisposableVMs</string>
      <x>0</x>
      <y>0</y>
      <width>1100</width>
-     <height>28</height>
+     <height>23</height>
     </rect>
    </property>
    <property name="contextMenuPolicy">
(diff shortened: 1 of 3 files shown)


andrewdavidwong:
I suggest keeping the orthography of `DisposableVM` and `DisposableVM Template` consistent, i.e., not:

- `Disposable VM` (extra space)
 - `Disposable VM Template` (extra space)
 - `DisposableVM template` (lowercase "T")

ui/qubemanager.ui
```diff
@@ -867,10 +867,10 @@ DisposableVMs</string>
     <bool>true</bool>
    </property>
    <property name="text">
-    <string>Is Template for DisposableVMs</string>
+    <string>DisposableVM Template</string>
    </property>
    <property name="toolTip">
-    <string>Is Template for DisposableVMs</string>
+    <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Disposable VM template&lt;/p&gt;&lt;p&gt;&lt;br/&gt;&lt;/p&gt;&lt;p&gt;Allows using this qube as a template for Disposable VMs. The Disposable VMs will inherit the VM's state (configuration, installed programs etc.), but their state will not be persistent. &lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
andrewdavidwong: > but their state will not be persistent.

I wonder if it would be clearer to say that they will be permanently destroyed when the initial window is closed. I wonder if there's a more intuitive way to say "the initial window."

Any ideas, @marmarek?

marmarta: Yeahhh, let's not get into service calls when non-terminal users can see it :) Maybe "their state will not persist between restarts"?

andrewdavidwong: > Maybe "their state will not persist between restarts"?

Sounds good.

marmarta:
I fixed the issues discussed and pushed a fixed version; @marmarek, it's ready to merge.

PR #192, opened 2019-08-08
Added virt mode column to Qube Manager

marmarta:
Just another column that shows the
contents of virt_mode property.

fixes QubesOS/qubes-issues#4702


marmarta:
I have a handler for property-set:*, that does include virt-mode

marmarek:
Ah, indeed.

PR #193, opened 2019-08-09
Fixed bug with port numbers and services in VM settings

marmarta:
Opening an existing FW rule resulted in an invalid port number.


qubesmanager/firewall.py
```diff
@@ -146,6 +146,7 @@ def __init__(self, parent=None):
         self.__services = list()
 
         self.port_range_pattern = re.compile(r'\d+-\d+')
+        self.service_port_pattern = re.compile(r'(\d*) \([a-zA-Z-]*\)')
```
marmarek: Service name can contain digits too. I'll fix it in a separate commit.

PR #195, opened 2019-08-25
Migration to View/Model design

donob4n:
Hi,

I started to work on a migration to View/Model concept for Qubes Manager. It is in early stage yet but I hope I could finish in few weeks and I want to hear some opinions and comments, also give it some visibility in case another dev starts another pretty big rewrite.

Currently no action is working, it just displays info.

### A brief  explanation
View/Model tries to maintain separated the data from the code which handles its representation and decoration. In this case, we could just remap the data from a VM object, 'column 2 -> self.vm.name e.g.', but this causes too much overload on qubesd, so I use a class for caching all needed data (VmInfo). The model (QubesTableModel) fills a list of VmInfo objects and it's ready for start drawing the table. 

The current version already does some kind of caching, storing needed values on the widgets for speed up sorting. With this version all the cached info is in the same place and there is no duplication.

The other part of the model is on data() method. Here we handle the information displayed and the decoration, also we could handle user input like editing directly a cell of the table.

I am pretty happy with the changes. Although the appearance is not yet identical, it is pretty close and 'table_widgets.py', a place with a lot of complexity, is fully removed. I will restore some code from there, but I don't think a separate file will be needed. 

The main objective is to have more readable code and easier to maintain. Also once it's finished amazing features will be pretty simple to add (multiple selection for manage multiple VM's at once, changing options like template/netvm directly on the table with a ComboBox, change table to a tree and have VM's sorted in different groups...). However it also seems to improve performance. Startup process has a pretty big boost, I get from 3.92s to 2s :)


marmarek:
Heads up: I've just merged #178 which most likely conflicts with this one. Also, `master` branch now is about R4.1 version (as both this and migration to Qt5 are unsuitable for stable updates). In practice it still should be possible to test it on R4.0, but it isn't guaranteed to work anymore.

As for the actual change, generally sounds good (I haven't looked into details, as some of it isn't finished yet). Please keep precise data updates working - I mean if some property is changed (and you get an event about it), refresh only that property, not the whole VM.

donob4n:
No problems with that @marmarek . 

I am working on it but unfortunately I am having problems with 'State' column and his multiple icons/tooltips. I asked for some help at pyqt mail list, I hope to solve it soon.

donob4n:
Still working on this, unfortunately I had pretty difficult days and I could not work on this.

I merged with PyQt5 version, solved my problem with multiple icons in a same cell (thanks to pyqt guys), and started the multi-row support.

donob4n:
Hi, 
This is nearly finished but I have some problems and questions:
- When opening settings windows I get this error:

```
  File "/usr/lib/python3.5/site-packages/qubesmanager/utils.py", line 186, in prepare_kernel_choice
    kernels = [kernel.vid for kernel in app.pools['linux-kernel'].volumes]
  File "/usr/lib/python3.5/site-packages/qubesmanager/utils.py", line 186, in <listcomp>
    kernels = [kernel.vid for kernel in app.pools['linux-kernel'].volumes]
  File "/usr/lib/python3.5/site-packages/qubesadmin/storage.py", line 327, in volumes
    'dom0', 'admin.pool.volume.List', self.name, None)
  File "/usr/lib/python3.5/site-packages/qubesadmin/app.py", line 577, in qubesd_call
    return self._parse_qubesd_response(return_data)
  File "/usr/lib/python3.5/site-packages/qubesadmin/base.py", line 81, in _parse_qubesd_response
    'Got empty response from qubesd. See journalctl in dom0 for '
qubesadmin.exc.QubesDaemonNoResponseError: Got empty response from qubesd. See journalctl in dom0 for details.
Aborted (core dumped)
```
Is this due some change in qubesadmin? I also got a pylint error about no 'ecx.QubesNoSuchPropertyError'

Maybe I have to test it on 4.1.

- I would like to add a 'template change' option in context menu. It will make 'Template Manager' unnecessary since you will can edit many VM's at once with multiple row selection. Are you ok with it? Do you prefer using this PR or better wait for another?

- I also would like to add some filtering options like 'Running domains', show only some Labels, show/hide sys VM's... what do you think? Use this PR also?

I would like to hear some comments (@marmarek , @marmarta) and what do you think about adding more things or merge first properly first.

marmarek:
> Is this due some change in qubesadmin? 

Looks like you run it on R4.0. This part needs a commit that is in R4.1 only: https://github.com/QubesOS/qubes-core-admin/commit/263f218d40fb93b97efacacc75808183f5e4dfb3#diff-0c889398bb7fbe645fcda185af7132de

> I also got a pylint error about no 'ecx.QubesNoSuchPropertyError'.

A typo? Should be exc, not ecx.

>     * I also would like to add some filtering options like 'Running domains', show only some Labels, show/hide sys VM's... what do you think? Use this PR also?

Better separate PR, there is already enough changes here.

>     * I would like to add a 'template change' option in context menu. It will make 'Template Manager' unnecessary since you will can edit many VM's at once with multiple row selection. Are you ok with it? Do you prefer using this PR or better wait for another?

I don't think changing template should be too easy. This can easily lead to breaking your system by a single mis-click. Leaving alone the fact you can't change the template of a running VM.
Anyway, lets discuss it outside of this PR.

donob4n:
> > I also got a pylint error about no 'ecx.QubesNoSuchPropertyError'.
> 
> A typo? Should be exc, not ecx.
> 

Sorry I miss typed here. I tried again with travis and the error appears again:
```
************* Module qubesmanager.qube_manager

qubesmanager/qube_manager.py:237:23: E1101: Module 'qubesadmin.exc' has no 'QubesNoSuchPropertyError' member (no-member)
```

> 
> Better separate PR, there is already enough changes here.
> 

Ok last thing I want to do is avoid "columns_indices" redundancy.  I just discovered what 'columns_actions', I think that I could rewrite the view/hide columns in a easier way.

marmarek:
> qubesmanager/qube_manager.py:237:23: E1101: Module 'qubesadmin.exc' has no 'QubesNoSuchPropertyError' member (no-member)

That's because for pylint purposes we use mockup of `qubesadmin.exc` module in https://github.com/QubesOS/qubes-manager/blob/master/test-packages/qubesadmin/exc.py. Copying needed exception class from original `qubesadmin/exc.py` should fix this.

donob4n:
> > qubesmanager/qube_manager.py:237:23: E1101: Module 'qubesadmin.exc' has no 'QubesNoSuchPropertyError' member (no-member)
> 
> That's because for pylint purposes we use mockup of `qubesadmin.exc` module in https://github.com/QubesOS/qubes-manager/blob/master/test-packages/qubesadmin/exc.py. Copying needed exception class from original `qubesadmin/exc.py` should fix this.

Ah, ok. Thanks Marek.

qubesmanager/qube_manager.py
```diff
@@ -61,140 +70,363 @@ def mousePressEvent(self, e):  # pylint: disable=invalid-name
...
+                self.state['outdated'] = 'update'
+
             if not event or event.endswith(':label'):
-                self.label_widget.update()
+                self.label = self.vm.label
             if not event or event.endswith(':template'):
-                self.template_widget.update()
+                try:
+                    self.template = self.vm.template.name
+                except exc.QubesNoSuchPropertyError:
```
marmarek, line 238: This should be AttributeError (exc.QubesNoSuchPropertyError inherits from it, so it will be matched too)

qubesmanager/qube_manager.py
```diff
@@ -61,140 +70,363 @@ def mousePressEvent(self, e):  # pylint: disable=invalid-name
...
+                    self.template = self.vm.klass
```
marmarek, line 239: This should be somehow differentiated, as it isn't really a template. Old version has it grayed out.

qubesmanager/qube_manager.py
```diff
@@ -61,140 +70,363 @@ def mousePressEvent(self, e):  # pylint: disable=invalid-name
...
+                if self.vm.features.get('internal', False):
+                    self.internal = "Yes"
+                else:
+                    self.internal = ""
             if not event or event.endswith(':ip'):
-                self.ip_widget.update()
+                self.ip = getattr(self.vm, 'ip', None)
             if not event or event.endswith(':include_in_backups'):
-                self.include_in_backups_widget.update()
+                self.inc_backup = getattr(self.vm, 'include_in_backups', None)
```
marmarek, line 253: Bool properties better present as `Yes` / empty, as you've preserved it in `internal` property.

qubesmanager/qube_manager.py
```diff
@@ -61,140 +70,363 @@ def mousePressEvent(self, e):  # pylint: disable=invalid-name
...
+                        self.last_backup))
             if not event or event.endswith(':default_dispvm'):
-                self.dvm_template_widget.update()
+                self.dvm = getattr(self.vm, 'default_dispvm', None)
             if not event or event.endswith(':template_for_dispvms'):
-                self.is_dispvm_template_widget.update()
-            if not event or event.endswith(':virt_mode'):
-                self.virt_mode_widget.update()
+                self.dvm_template = getattr(self.vm, 'template_for_dispvms',
+                                            None)
```
marmarek, line 263: Same here - bool property.

qubesmanager/qube_manager.py
```diff
@@ -634,37 +812,39 @@ def on_domain_changed(self, vm, event, **_kwargs):
...
-        try:
-            self.table.sortItems(self.columns_indices[self.sort_by_column],
-                                 self.sort_order)
-        except KeyError:
-            # the manager was sorted on a column that does not exist in the
-            # current version; possible only with downgrades
-            self.table.sortItems(self.columns_indices["Name"],
-                                 self.sort_order)
+        # Restore sorting
+        sort_column = int(self.manager_settings.value("view/sort_column"))
```
marmarek, line 841: This (and some other places in this function) fails if setting wasn't there  (`int(None)`), use `defaultValue=` argument.

qubesmanager/qube_manager.py
```diff
@@ -634,37 +812,39 @@ def on_domain_changed(self, vm, event, **_kwargs):
             return
 
         try:
-            self.vms_in_table[vm.qid].update(event=event)
+            self.qubes_cache.get_vm(qid=vm.qid).update(event=event)
+            self.proxy.invalidate()
         except exc.QubesPropertyAccessError:
             return  # the VM was deleted before its status could be updated
 
     def load_manager_settings(self):
```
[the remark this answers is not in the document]
donob4n: Yes I did there because it sets all checked or unchecked options but it would be better to init the menu first.

qubesmanager/qube_manager.py
```diff
@@ -1209,50 +1313,6 @@ def showhide_toolbar(self, checked):
     def showhide_column(self, col_num, show):
         self.table.setColumnHidden(col_num, not show)
 
-        if self.settings_loaded:
-            col_name = [name for name in self.columns_indices if
-                        self.columns_indices[name] == col_num][0]
-            self.manager_settings.setValue('columns/%s' % col_name, show)
-            self.manager_settings.sync()
```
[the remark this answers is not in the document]
donob4n: Ouch, I tougth that closeEvent was always called. I will restore it.

donob4n:
> Generally it is a big cleanup, thanks!
> 
> Besides minor comments below, at startup it shows `QPixmap::scaled: Pixmap is a null pixmap`, I haven't tracked it down.

The _null pixmap_'s are dom0 and disp-vm's, I will check the icons that it tries to load compared to old version.

donob4n:
I've noticed it says directly default netvm instead "default (sys-firewall)", I guess it is important information but I see the final line too long. What about putting just "default" and add the current default as a tooltip?

marmarek:
I'd rather put the actual value, and mark being default somehow else (italic + tooltip?) @marmarta

marmarta:
> I'd rather put the actual value, and mark being default somehow else (italic + tooltip?) @marmarta

I'd rather keep the default (actual value) here... I know it's long, but both points of data are important, and tooltips have the problem of just displaying one at a time.

donob4n:
Well, the default value is the same for all domains using it, you just need to show the tooltip once.

marmarta:
> Well, the default value is the same for all domains using it, you just need to show the tooltip once.

Yeah, but this is not the rule with all defaults: from what I understand, some properties might have different defaults according to things like 'is this a hvm' or 'does it have PCI devices'. It's may not be super obvious which properties are which.

marmarek:
I've tested the current version (https://github.com/QubesOS/qubes-manager/pull/195/commits/a1d96e78778c84fe077b62196c8ed561978de9bc) and I think handling default value discussed above is the only remaining issue, other things are fine and good for merge.
There is a merge conflict here, but it should be easy to resolve (please do that when you push changes, so tests will run automatically).

qubesmanager/qube_manager.py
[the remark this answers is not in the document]
donob4n: Ey @marmarek , according to QT help:

> void QSettings::sync() 
> Writes any unsaved changes to permanent storage, and reloads any settings that have been changed in the meantime by another application. This function is called automatically from QSettings's destructor and by the event loop at regular intervals, so you normally don't need to call it yourself.

So I did some testing after removing direct calls to it. It works fine even killing the process. I guess it should only fail in some strange case when the process dies at the exact moment of doing a sort change or other setting.  The real problem was closeEvent() not being called on forced termination, so I moved the settings.setValues() calls to layoutChanged and resizeEvent().

donob4n:
I was trying to resolve the conflicts but I have to let it for tomorrow. I'm getting empty responses from from qubesdb when I open a settings dialog.

donob4n:
It was the problem with kernel list problem with 4.0, there is another new incompatibility with 'qvm-appmenus' and '--template' argument.

At this point having compatibility with 4.0 is very simple at this point (maybe better to wait for adding some additional features). I think you could consider it, 4.0 will survive antoher a year or more, right?

marmarek:
> At this point backporting to 4.0 is very simple

It isn't only whether it's simple or not. R4.0 is a stable release, where the priority is - stability, not new features (some details [here](https://www.qubes-os.org/doc/version-scheme/)). New features, refactors etc do include risk of regression, which we want to avoid in a stable branch.

marmarek:
Ok, the code looks good now!
Can you also update the tests (`test_qube_manager.py` file)? Unfortunately this looks like a major change there. If you need any help @marmarta will be happy to assist.

donob4n:
Oh yes, I will try to fix it.

donob4n:
Is there some easy way for debug it? I added some output but I can't see anything when running tests.

marmarta:
Normal print works for me - are you certain you are editing the correct files? The command @marmarek  gave is for running tests on an installed package, not in the source (this is something that should be done one day, making good tests that do not rely on everything else... but it's not done yet :( )

marmarta:
Where correct packages are in somewhere like `/usr/lib/python3.8/site-packages/qubesmanager/tests/test_qube_manager.py` (or python3.7, depending on the version of everything else you have)

donob4n:
Ohh thanks @marmarta , it seems that I added debug inside '_get_table_item()' but it is crashing before reaching there. All under control now :)

marmarek:
@donob4n how is adjusting tests going? I refrain from merging other pull requests here to avoid conflicts, but there are more and more of them. Do you need some help?

donob4n:
Wops I've fixed many but still having 48 errors. I will try to fix more tonight and at least commit something.

donob4n:
Now I can't commit due my pgp key expired. I hope just creating a newer one should work.

donob4n:
@marmarek I edited the key expiration and sent it to pgp.mit.edu

Still 17 failures and 24 errors, I wil try to fix more tomorrow and comment some that seem pretty difficult to fix.

donob4n:
failures=12, errors=5, expected failures=1

I hope to finish it tomorrow.

donob4n:
Ey  @marmarek , @marmarta this is near ready. I have a problem with 'test_218_remove_vm_dependecies', "AssertionError: Expected call: show()"

Also sometimes 'test_500_logs' fails but I can't reproduce it always.

marmarta:
@donob4n awesome! test_500_logs used to be flaky sometimes, so I'm going to assume we stumbled on yet another Qt weirdness. I'll check out the test_218 bug, but I'm guessing it's a test bug, not a program bug :) 
@marmarek , I think we should be merging this soon :)

qubesmanager/qube_manager.py
```diff
@@ -61,140 +71,397 @@ def mousePressEvent(self, e):  # pylint: disable=invalid-name
...
+                ret = vm.label.name
+            elif col_name == "State":
+                ret = str(vm.state)
+            elif col_name == "Disk Usage":
+                ret = vm.disk_float
+            else:
+                ret = self.data(index, Qt.DisplayRole)
+
+            if col_name != "Name":
+                ret = ret + vm.name
```
marmarek, line 442: The tests checks if dom0 is sorted first, but this doesn't guarantee it. In fact, many of those tests fails for me. For example when sorting by "Template", it gets `dom0` vs `debian-10-xfcesys-gui` and the latter is sorted earlier.

marmarek:
>  I have a problem with 'test_218_remove_vm_dependecies', "AssertionError: Expected call: show()"

That's because you've changed how QMessageBox is imported. When you do `from ... import QMessageBox`, you need to mock `qubesmanager.qube_manager.QMessageBox`.

qubesmanager/qube_manager.py
```diff
@@ -61,140 +71,397 @@ def mousePressEvent(self, e):  # pylint: disable=invalid-name
...
-                self.last_backup_widget.update()
+                self.last_backup = getattr(self.vm, 'backup_timestamp', None)
+                if self.last_backup:
+                    self.last_backup = str(datetime.fromtimestamp(
+                        self.last_backup))
             if not event or event.endswith(':default_dispvm'):
-                self.dvm_template_widget.update()
+                self.dvm = getattr(self.vm, 'default_dispvm', None)
+                if self.vm.property_is_default("default_dispvm"):
+                    self.dvm = "default (" + self.dvm.name + ")"
```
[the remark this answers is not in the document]
marmarek: Hmm, `self.vm` is a `VmInfo` object, not `QubesVM`, right? Then, this would require adding `__str__` method.

BTW, in the meantime `QubesVM` object in `qubesadmin` module got caching support directly (including reacting to events to update it), so you can use that directly now.

qubesmanager/qube_manager.py
donob4n: Oh, it was ' return "" ' for dom0 and worked fine but I changed it fixing a pylint warning :rofl:

donob4n:
> That's because you've changed how QMessageBox is imported. When you do `from ... import QMessageBox`, you need to mock `qubesmanager.qube_manager.QMessageBox`.

I think that the problem is that there is no row selected so 'action' is not enabled and trigger() does nothing.  I would try to select the 'test-vm' row but I don't understand this lines:

`
mock_vm = unittest.mock.Mock(spec=['name'],
                                     **{'name.return_value': 'test-vm'})
mock_dependencies.return_value = [(mock_vm, "test_prop")]
`

I manually created both 'test-vm' (providing network) and 'test_prop' with test-vm as net-vm. Should be enough to select test-vm before calling trigger()?

marmarek:
> I think that the problem is that there is no row selected so 'action' is not enabled and trigger() does nothing. I would try to select the 'test-vm' row but I don't understand this lines:

Yes, that's indeed the other thing there. Just select some not running VM.

> I would try to select the 'test-vm' row but I don't understand this lines:

No need - it is mock return value what _other_ VMs depends on the one you're going to remove. Since it's mockup, it doesn't actually need to exist at all.

qubesmanager/qube_manager.py
marmarek: This concatenation has other unintended effects:
```
======================================================================
FAIL: test_100_sorting (qubesmanager.tests.test_qube_manager.QubeManagerTest)
----------------------------------------------------------------------
  File "/usr/lib/python3.8/site-packages/qubesmanager/tests/test_qube_manager.py", line 251, in test_100_sorting
    self.__check_sorting("Template")
  File "/usr/lib/python3.8/site-packages/qubesmanager/tests/test_qube_manager.py", line 1266, in __check_sorting
    self.assertGreater(
AssertionError: 'fedora-32' not greater than 'fedora-32-xfce' : Incorrect sorting for Template

----------------------------------------------------------------------
```

That's because it isn't `fedora-32` sorted against `fedora-32-xfce` - it is `fedora-32vmname` sorted against `fedora-32-xfcevmname` and indeed this reverses the order here. Adding some separator that cannot appear in the column content (`!`? a space?) should help, but still feels hacky. Perhaps use tuples (primary value, vmname) to make sure the vm name is used for sorting really only when the first value matches?

donob4n: I tried with a better way, or a least not so hacky.

marmarek:
> That's because you've changed how QMessageBox is imported. When you do `from ... import QMessageBox`, you need to mock `qubesmanager.qube_manager.QMessageBox`.

This applies to other tests too, specifically VMShutdownMonitorTest class.

marmarek:
Merged by squashing all the commits into one to limit noise in the git history.
Thanks @donob4n !

donob4n:
Ohh great!

donob4n:
Ey @marmarek , I would like to fix some sorting behaviour and to add filtering options but my fork repo is 107 commits ahead.

I suppose that it's due squashing  but I don't know what should I do for fix it.

marmarek:
If you didn't made any other changes in the repo, you can reset your repo to the upstream state, like this:
```
git fetch https://github.com/qubesos/qubes-manager
git reset --hard FETCH_HEAD
git push -f
```

For the next time, I strongly recommend doing any changes in a separate branch, not master.
BTW there are other PRs open in this repo fixing similar issues. I'm waiting for the test run to finish (probably 2h more) and will merge them.

donob4n:
Ok thanks!

PR #198, opened 2019-09-12
Fixed new firewall rule window

marmarta:
Now, an error will not close the window, enabling
the user to correct any mistakes.

fixes QubesOS/qubes-issues#bugfix5234


qubesmanager/firewall.py
```diff
@@ -145,6 +206,9 @@ class QubesFirewallRulesModel(QtCore.QAbstractItemModel):
     def __init__(self, parent=None):
         QtCore.QAbstractItemModel.__init__(self, parent)
 
+        self.current_row = None
+        self.current_dialog = None
```
marmarek: This field is unused, please remove.

marmarta: pushed a fix

PR #201, opened 2019-09-27
Show the previous VM name while renaming

SaswatPadhi:
Currently while renaming VMs, the default text is empty "". This change modifies this default text to the original name of the VM.

This makes it a lot easier to make small changes to the name. The original name is also _selected_ by default, so replacing it entirely is also easily -- simply start typing the new name.


SaswatPadhi:
Looks like the build failure is because of `pylint` warnings, and should be resolved after PR #200 has been merged ...

PR #205, opened 2019-10-16
Manager tests fixed for Qt5

marmarta:
A bunch of fixes for Qt5, including some fixes for features introduced after the switch.
Also a minor fix for issue found by tests.
And a fix for unhandled exception in backup.reject method that caused segfault when the backup
tool was ran without registering additional exception handler.

fixes QubesOS/qubes-issues#5395


marmarek:
Note from testing it on OpenQA: when test_backup is _not_ called first, it crashes on `test_20_loading_settings` test. Apparently it's about usage of `self.dialog.deleteLater()`. Adding `self.qtapp.processEvent()` after every `self.dialog.deleteLater()` helps.

PR #206, opened 2019-10-18
Allow build for VM

fepitre:
QubesOS/qubes-issues#833
QubesOS/qubes-issues#5403


marmarek:
Please don't forget about adding issue reference for tracking purposes. I've edited description here.

Makefile.builder
```diff
@@ -1,3 +1 @@
-ifeq ($(PACKAGE_SET),dom0)
```
marmarek: Please consider all supported rpm-based distributions. Specifically I think this will fail to on CentOS. If that's true, it's fine to exclude centos build for now. But it is not ok to have enabled package that doesn't build.

PR #210, opened 2019-10-28
Translation update

marmarta:
Translation files updated for Qt5. 

references QubesOS/qubes-issues#5410


qubesmanager/appmenu_select.py
```diff
@@ -85,6 +84,8 @@ def save_appmenu_select_changes(self):
             stdin=subprocess.PIPE)
         p.communicate('\n'.join(new_whitelisted).encode())
         if p.returncode != 0:
-            raise RuntimeError('qvm-appmenus --set-whitelist failed')
+            exception_text = QtCore.QCoreApplication.translate(
+                "exception", 'qvm-appmenus --set-whitelist failed')
```
marmarek: Better use format like `Command '{}' failed` and fill the command outside of translated string. Will be easier to translate, and also reusable in other messages.

qubesmanager/global_settings.py
```diff
@@ -400,7 +408,9 @@ def save_and_apply(self):
 
 
 def main():
-    utils.run_synchronous("Qubes Global Settings", GlobalSettingsWindow)
+    utils.run_synchronous(
+        QtCore.QCoreApplication.translate("appname", "Qubes Global Settings"),
```
marmarek: This won't work - translator is not initialized yet.

qubesmanager/settings.py
```diff
@@ -1213,8 +1211,9 @@ def main(args=None):
     args = parser.parse_args(args)
     vm = args.domains.pop()
 
-    utils.run_synchronous("Qube Settings",
-                          functools.partial(VMSettingsWindow, vm, args.tab))
+    utils.run_synchronous(
+        QtCore.QCoreApplication.translate("appname", "Qube Settings"),
```
marmarek: This won't work - translator is not initialized yet.

qubesmanager/template_manager.py
```diff
@@ -378,9 +381,10 @@ def vm_state_change(self, is_running, row=None):
 
 
 def main():
-    utils.run_asynchronous("Template Manager",
-                           "qubes-manager",
-                           TemplateManagerWindow)
+    utils.run_asynchronous(
+        QtCore.QCoreApplication.translate("appname", "Template Manager"),
```
marmarek: This won't work - translator is not initialized yet.

PR #214, opened 2019-11-10
Silence an unnecessary warning in the "Backup Qubes" gui

hexagonrecursion:
https://github.com/QubesOS/qubes-issues/issues/5427


marmarta:
I don't like the total loss of information with regards to some config data not being loaded. Maybe add a less-scary label to show only in those cases (and load default)? Or ask the user if the VM should be started (if it's not running).

marmarta:
I've fixed the same issue it in a less-information losing form here: https://github.com/QubesOS/qubes-manager/pull/254 . Would propose closing this PR.

PR #215, opened 2019-11-11
Fix locale

fepitre:


qubesmanager/global_settings.py
```diff
@@ -32,14 +33,20 @@
 
 qmemman_config_path = '/etc/qubes/qmemman.conf'
 
+
 def _run_qrexec_repo(service, arg=''):
+    # Set default locale to C in order to prevent error msg
+    # in subprocess call related to falling back to C locale
+    env = os.environ
```
marmarek:
```suggestion
    env = os.environ.copy()
```
Otherwise you'll modify global instance.

PR #236, opened 2020-05-15
Better error handling for opening VM settings in Qube Manager

marmarta:
No more silent errors - the user will get an error message.

fixes QubesOS/qubes-issues#5810

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -1042,43 +1042,51 @@ class VmManagerWindow(ui_qubemanager.Ui_VmManagerWindow, QtWidgets.QMainWindow):
                         "ERROR: {1}").format(vm.name, ex))
                 return
 
+    def open_settings(self, vm, tab='basic'):
+        try:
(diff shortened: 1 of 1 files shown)


qubesmanager/qube_manager.py
```diff
@@ -1042,43 +1042,50 @@ def action_killvm_triggered(self):
...
+            # have been cloned or removed
+            self.vms_in_table[vm.qid].update()
+        except exc.QubesException as ex:
+            QtWidgets.QMessageBox.warning(
+                self,
+                self.tr("Qube settings unavailable"),
+                self.tr(
+                    "Qube settings cannot be opened. The qube might have "
+                    "been removed or unavailable due to policy settings."
+                    "\nError: {}".format(str(ex))))
```
marmarek, line 1064: Does this cover the case of removing/renaming VM (from within the settings window)? If so, it isn't an error. I think you need to catch an error and show this message if earlier part of this function fails (either constructing the object, or exec_() call - not sure.

PR #238, opened 2020-05-28
Fixed VM settings test

marmarta:
Incorrect maxmem value caused error.

fixes QubesOS/qubes-issues#5857


marmarta:
@marmarek , something exploded all over here

PR #239, opened 2020-06-02
 Added a Clone Qube tool

marmarta:
A simple GUI tool to enable cloning a VM to a different pool and changing the label. Qube Settings and Qube Manager now use it, too.

fixes QubesOS/qubes-issues#5127
fixes QubesOS/qubes-issues#5978


marmarta:
@marmarek , rebased onto new master

PR #250, opened 2020-07-21
fixed misspelling of success in dialog box

hexparrot:


hexparrot:
forgot to add signed pgp key

PR #254, opened 2020-08-06
Modified warning shown when destination qube for backup is unavailable

marmarta:
Instead of scary red warning, now a gentler information that only running qubes are
available is shown. It is also shown just under the destination VM dropdown,
making it easier to notice and interpret correctly.

fixes QubesOS/qubes-issues#5427


marmarta:
note: the failed test above is due to connection problems, not code problems.

PR #255, opened 2020-08-07
shutdown-when-idle gets its own checkbox in VM settings

marmarta:
It is disabled if the service is unavailable and displays information
about how to enable it if it is unavailable.

references QubesOS/qubes-issues#5298

### DIFF
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -438,6 +438,31 @@ class VMSettingsWindow(ui_settingsdlg.Ui_SettingsDialog, QtWidgets.QDialog):
             self.include_in_backups.setEnabled(False)
 
         try:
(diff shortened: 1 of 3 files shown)


marmarek:
>     * qubesmanager_vmsettings: [Failed](https://openqa.qubes-os.org/tests/11908#step/qubesmanager_vmsettings/8) (`# Test died: no candidate needle with tag(s) 'vm-settings-correct_current' matched`)

The "unavailable" version of the checkbox text makes the settings window overly wide, missing word wrap enabling?

marmarek (changes requested):
The "unavailable" version of the checkbox text makes the settings window overly wide, missing word wrap enabling?

marmarek (approved):

PR #256, opened 2020-08-11
Modified Qube Manager toolbar to be more readable and added Emergency Pause

marmarta:
Instead of tiny buttons, now Qube Manager will have bigger buttons and text underneath.
Furthermore, Pause gets renamed to Emergency pause for clearer communication.

references QubesOS/qubes-issues#5993


marmarta:
This is a single PR and not two because of the nightmare that is resolving merge conflicts in .ui files.

marmarek (approved):

PR #258, opened 2020-08-18
qvm-template GUI

WillyPillow:


marmarek:
The rest of the GUI tools here are in PyQt, so it would be much more consistent to write this one in PyQt too...

WillyPillow:
Oops. For some reason, I've always assumed in previous discussions that PyGTK is preferred, perhaps as it's listed as a prerequisite on the GSoC page. In retrospect, I should have made sure to ask explicitly (actually thought I did but apparently not). Sorry for the blunder.

I should be able to port the current code to PyQt in a reasonable amount of time. But in the meantime, are there any high-level issues with the current design?

marmarek:
Generally it looks fine. Some ideas for improvement (that would apply regardless of the graphical toolkit):
  - since we use asyncio, it makes more sense to use that to run qvm-template command without blocking GUI, instead of separate thread
  - create UI layout in an appropriate tool (for qt it would be qt designer, for gtk - glade) - this makes it easier to maintain the code, but also you get translation support for free

marmarek:
I've checked the code (not very thoroughly yet) and it looks good, but I cannot test it right now. Can you post a screenshot here how the layout looks in practice?

WillyPillow:
(Since the second shot was taken afterwards, it says "remove" instead of "install".)

marmarta (changes requested):
Looks good! There's a minor packaging issue - at the moment, manager won't build. It needs adding the four files you add to rpm_spec/qmgr.spec.in (for fedora) and debian/install (for debian).

You can also add an entry point in setup.py (let me propose qvm-template-gui for the moment) (the resulting binary also has to be added to spec/install files).

marmarta:
One more thing: I think size would be more readable in MB, not in kB. The large numbers are a bit hard to parse visually - and as far as template sizes go, MB is the more typical unit.

WillyPillow:
> One more thing: I think size would be more readable in MB, not in kB. The large numbers are a bit hard to parse visually - and as far as template sizes go, MB is the more typical unit.

Done.

> Looks good! There's a minor packaging issue - at the moment, manager won't build. It needs adding the four files you add to rpm_spec/qmgr.spec.in (for fedora) and debian/install (for debian).
> 
> You can also add an entry point in setup.py (let me propose qvm-template-gui for the moment) (the resulting binary also has to be added to spec/install files).

I'll work on this soon. Speaking of naming, as currently this and the existing template manager are both presented to the user as "Template Manager", would it make sense to tweak the names of one or the other to reduce confusion?

marmarta:
> I'll work on this soon. Speaking of naming, as currently this and the existing template manager are both presented to the user as "Template Manager", would it make sense to tweak the names of one or the other to reduce confusion?

I'll try to merge them into one tool sometime in September (or, to be exact, I'll probably extend your tool with, because it's written much more nicely). Hm, maybe until a merge comes, we could call this one qubes-template-installer and the other qubes-template-manager ?

WillyPillow:
I see. Merging/integration with the current template manager is actually one of the initial nice-to-have goals of the project. I'd love to help work on this, but currently my focus is on writing test cases for <https://github.com/QubesOS/qubes-core-admin-client/pull/145>.

I'm guessing that even when they're merged, there still need to be some distinction between the two components though.

marmarek:
PipelineRetry

marmarek:
PipelineRetry

marmarek:
PipelineRetry

ninavizz:
- **Super nitty feedback:**
  - `Reponame` should be two words, Repo Name. Non-developers will google the single word and be sad.
  -  I would remove the word "Time" from the Build/Remove headers. It's clear they're timestamps, yet they begin with the date and it's strange seeing the word "Time" and then a bunch of dates. TMI = cognitive burden. Remove the burden. :)
  - What value does licensing information have to anybody? If the answer is "not much" I'd prefer it not get shown. Again, cognitive burden. If I'm stealing software, I'll know I'm stealing it—so why remind me?
- **Less Nitty Feedbck**
  - The presentation of the date/time stamps is not very user-friendly. What is most important to the user, the day, the year, or the time? Since it's truncated, that matters. Likewise, dates should show as `28 Aug 2000` vs `2000-08-28`. If clock-time matters more, I'd say format as `23:42, 28 Aug 2000` but if the date is what matters the most, I'd reverse that.
  - I realize it's the norm to include distro names in the name of the Template itself, but that means reeeeeally long names sometimes. It seems like it'd be helpful to have a column to show the distribution name, before the version number. Like, if I'm a super Ubuntu user but don't want 10,000 apps installing on all my qubes, I'll have 3 templates with 3,300 apps apiece on them—and it'd be nice to just say `Personal` or `Work` or `Art` than `Ubuntu-18-personal`. We're currently forcing users into a naming convention, and it'd be nice to not have to do that. 

I still don't know how to create a new template, on an aside. It feels like an absent feature in this? That said, I don't want `4.1` held-up any longer, and would only request my nitty feedback points be considered, immediately.

marmarek:
> I still don't know how to create a new template, on an aside. It feels like an absent feature in this?

That was my first impression too, honestly. There was a help test (that you need to click "status" column), but it's visible only before selecting anything. I've already added a persistent help text, updated screenshot:

qvm-template-gui

ninavizz:
> That was my first impression too, honestly. There was a help test (that you need to click "status" column), but it's visible only before selecting anything. I've already added a persistent help text, updated screenshot

That is certainly helpful—but  if I want to create a new Template, why would I click on an existing line item? That doesn't make sense. Honestly, I've already spent some time reading about creating an Ubuntu template, and I'm still in the dark about how. The docs assume some fluency with CLI stuff, and I feel like a tennis-ball being darted from article to article, there. Which I say with hesitance, as Andrew and Michael are so thoughtful with not wanting to ever leave users in the dark. 

While it's likely outside the scope of this issue, I'd love to see a "New Template" button and just be able to click it to get the ball rolling. To me this feels like a big "blocker" sort of a need for new Qubes non-techy folks.

marmarek:
I think I understand you now. Generally you don't _create_ new template. You download and install one from the repository. Creating new template is quite complex task that is definitely beyond end-user skill set ([docs](https://github.com/Qubes-Community/Contents/blob/master/docs/building/building-non-fedora-template.md)).
That said, there is an option to create new TemplateVM in a "new qube" dialog (in Qubes 4.1), which in practice is useful for creating Windows templates.

marmarek:
>     * Likewise, dates should show as `28 Aug 2000` vs `2000-08-28`

Honestly, I'm not really sure if that's a good idea. In case of build time, this date's purpose is:
 - see how fresh the template is (newer than you have installed? is it actively maintained?)
 - easily sort to have recent versions first (for this, you can sort by version number too)
So, it's more important to easily compare the dates, which IMO is easier to do with numbers.

But I agree the clock-time part is redundant.

marmarek:
>     * `Reponame` should be two words, Repo Name.

Maybe "Repository" would be clearer?

>     * Could the headers be a dark color with white text/arrows?

This comes from whatever theme you have set in your system - so it is consistent with other tools (including non-qubes ones). I don't think we should change this particular one, it's rather task for https://github.com/QubesOS/qubes-issues/issues/6414

>     * The line above the headers also adds to "noise" and seems like it should be removed

The one above the hint now? Yes, I think it can be removed. Technically I think it comes from a collapsed progress bar that appears there if you click "refresh".

ninavizz:
> The one above the hint now?

Now that the hint is there, it should probably stay. It just served no purpose with guiding content-consumption or page hierarchies, before. Now it does. 

> Maybe "Repository" would be clearer?

The word itself is not the point of disconnect; the concept, is. Nobody uses the word "Repositories" to speak to a code repo, so I don't feel it should be spelled-out. The problem is that non-technical users don't know what a code repo is—or, it's not their first association with the word "repository" or "repo." Like, especially in the US where everyone does everything with debt, a "repo" is when a debt collector comes to your house and takes your car away. There's a well known movie here called Repo Man, too. :)  

So: sit down and brace yourself. I had a conversation with @eloquence about this some time ago, with regards to technical skill of our users from the SecureDrop Workstation pilot... and he considers me more technically adept, than all but one of our Journalist users (that one person, being a former developer). He labeled me "a technical user." You and I both know I couldn't write a line of Python (or understand what somebody else's line of Python meant) if my life depended upon it. So, when we speak about "non-technical users," that's a good reference point to consider. It's a point I often come back to, myself.

What is the value of a code repo in the context of the task of creating a new Template? A user unfamiliar with how software is built, will have no idea. This is a big reason why Linux is so alienating to so many people. Similarly: to the user, they _are_ in fact creating a new template. The tbd template will in fact, be new to their system and new to them. Like, I want to get a new Mac, but I'm fine getting a 2020 model. To me, it'll still be a _new_ machine. Even if I get a reconditioned laptop.

"A new thing" is the non-technical user's own mental-model of the task, even though technically they're just installing an existing template package from a repo somewhere. On that point, I agree: the technical inaccuracy of a button labeled "Create New Template" would be bad. However, a button that said simply said "New Template" on it and then showed a dropdown when pressed, with three options (I think?) "Install from package," "Install from repo," or "Create my own" (or something like that) to initiate a task, would be both accurate, and not alienate new users.

How to be technically accurate w/o alienating non-technical and new users, is sincerely hard. But, it's not rocket science. We just can't get hung-up on existential depths of truth wrt technical accuracy, while also being mindful to not mislead technical users. It takes iteration and collaboration, is all... and sometimes this is also why icons can be better than words. :)

> Honestly, I'm not really sure if that's a good idea. In case of build time, this date's purpose is:
> 
> - see how fresh the template is (newer than you have installed? is it actively maintained?)
> - easily sort to have recent versions first (for this, you can sort by version number too)
> 
> So, it's more important to easily compare the dates, which IMO is easier to do with numbers.

Yes, numbers do make everything easier! The "problem" as the human user consuming the information though, is that everything is connected with hyphens, and everything's a number. So there is initially a cognitive burden of grokking "Ok, this is a date—there is the year, and that number there is higher than 12 so it's probably the date, and the other number is probably the month." 

When at least the month is spelled-out in alpha characters, it makes it clearer that the entire string is a date, and which is the month value and which is the day value. Maybe a second of mental processing time is spared, but across an experience those seconds add-up and contribute to cognition fatigue and just feeling drained, is all. 

On the point of assessing newness, when the first value is always the year, it's like reading the list of Qubes in today's app menu... you're reading the word "Domain" over and over again, and eventually just want to cut to the information that's of real value to you—where actual app-vm names are. Or in this use case, _what_ the signifiers of newness are. 

Bigger picture tho, this also kinda speaks to why I prefer view-filter panels instead of filtering on table headers: because the most user-friendly ways to display information for human consumption in a list view, rarely aligns with how information needs to exist in a table for headers to accurately change a sort-order.

marmarek:
> "A new thing" is the non-technical user's own mental-model of the task, even though technically they're just installing an existing template package from a repo somewhere. On that point, I agree: the technical inaccuracy of a button labeled "Create New Template" would be bad. 

Maybe "Add Template" then? But for this application specifically (which is specifically about downloading pre-built templates) "Install Template" is quite accurate and I think it should be also understandable - after all, this is what you do with a software - you don't "create Firefox", you "install Firefox". And then "a new thing" (an icon to launch Firefox) will appear in your computer.

Then, in some other place (when creating new qube? in Qubes Manager?) we could have "Add Template" button with options like "download and install" or "create from scratch (advanced)".

ninavizz:
> Maybe "Add Template" then? But for this application specifically (which is specifically about downloading pre-built templates) "Install Template" is quite accurate and I think it should be also understandable - after all, this is what you do with a software - you don't "create Firefox", you "install Firefox". And then "a new thing" (an icon to launch Firefox) will appear in your computer.

I really like "Add Template". It also aligns with many other interactions where you don't "create" a thing, but you do add a thing that already exists in a general sense, to a list of things that are only 'yours' by curation or access—such as bookmarks and playlists. "Install" only feels appropriate once a source/repo has been identified. It makes sense as a primary action button label once the repo has been selected, but not to lead-in to selecting the repo.

> Then, in some other place (when creating new qube? in Qubes Manager?) we could have "Add Template" button with options like "download and install" or "create from scratch (advanced)".

A Template is technically a qube, but to me feels bigger than _just_ a qube? I did look for a secondary-function within today's "Add Qube" window, so it's not impossible to discover there but it doesn't feel natural there. I realize Template Manager is intended as a maintenance tool, but honestly—that's where it seems like the most natural fit per expecteed user mental models. I can ask in our forthcoming testing sessions for the app menu?

WillyPillow:
> Yes, numbers do make everything easier! The "problem" as the human user consuming the information though, is that everything is connected with hyphens, and everything's a number. So there is initially a cognitive burden of grokking "Ok, this is a date—there is the year, and that number there is higher than 12 so it's probably the date, and the other number is probably the month."

Regarding the date issue, I would say it really depends on the culture. As someone with a Chinese background, anything *other* than `{YY,}YYMMDD` feels foreign. Perhaps it's better to use the user's locale settings for this? (As long as sorting is not affected, of course.)

ninavizz:
> Perhaps it's better to use the user's locale settings for this? (As long as sorting is not affected, of course.)

If it is an option to use the user's locale settings, yes, that would be ideal! If a combination of alpha/numeric characters would be an option, that'd also be best—but I appreciate that could make things very complicated.

12 August 2020 is a European standard, whereas August 12 2020 is an American standard; so 08-12-2020 could mean one of  two dates, depending on your locale and any back-end settings... whereas at least the month being spelled in alpha characters, would make it clearer. But I sense that would be a lot more work, and complicated by localisation needs.

Thx for the insight on your perspective as a Chinese cultured user! Always helpful to learn these things. :)

marmarek:
> Perhaps it's better to use the user's locale settings for this?

I've just tried. I've got `MM/DD/YYYY` format, which is the most confusing of all (sorry US people, but it _is_ very ambiguous, you basically broke `/` in any date format for the whole world). Even worse, when I change locale from `en_US` to `en_GB`, then I get `DD/MM/YY` (and yes, YY, not YYYY). And with `pl_PL` locale I get `DD.MM.YYYY`.
It is `%c` format (you can play with it using `LC_TIME=en_US.UTF-8 date +%c`).

Then, there is also full date+time format (`%x`), which is not what we want, but it does contain month spelled out, in most locales I've tried (but not in Chinese, I guess):
```
$ LC_TIME=en_US.UTF-8 date +%c
Sat 03 Apr 2021 02:53:02 PM CEST
$ LC_TIME=en_GB.UTF-8 date +%c
Sat 03 Apr 2021 14:53:07 CEST
$ LC_TIME=pl_PL.UTF-8 date +%c
sob, 3 kwi 2021, 14:53:10
$ LC_TIME=zh_CN.UTF-8 date +%c
2021年04月03日 星期六 14时55分51秒
```
But it's way too long for this column.

So, I guess if we really want month to be spelled out, we can't use locale-specific format as a whole, but to hardcode some specific format (and have just month name in the current language). Otherwise, there is [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601), which is "unambiguous and well-defined method of representing dates and times, so as to avoid misinterpretation of numeric representations of dates and times", specifically - YYYY-MM-DD for the date format.

ninavizz:
Dang @marmarek I can't believe you did all that exploration on this—thank you! Yeah, America has broken a whole lot for the entire world. Your quip to that effect did make me cackle very loudly in my partner's ear just now, and he fully agrees. TY!! :)

marmarek:
Thanks @ninavizz for the feedback! I've applied the easiest ones already. I'm going to merge it as it is now (as it blocks R4.1), we can track/discuss further improvements in separate ticket.

marmarek (approved):

PR #263, opened 2020-09-09
Warn if trying to change layout on unsupported vms in Qube Manager

marmarta:
requires https://github.com/QubesOS/qubes-core-admin-client/pull/161
references QubesOS/qubes-issues#6030

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -1318,7 +1318,17 @@ class VmManagerWindow(ui_qubemanager.Ui_VmManagerWindow, QMainWindow):
     def action_set_keyboard_layout_triggered(self):
         # pylint: disable=invalid-name
         for vm_info in self.get_selected_vms():
-            vm_info.vm.run('qubes-change-keyboard-layout')
+            if vm_info.vm.features.check_with_template(
+                    "supported-feature.keyboard-layout", False):
+                vm_info.vm.run('qubes-change-keyboard-layout')
+            else:
+                QMessageBox.warning(
+                    self,
+                    self.tr("Keyboard layout change unsupported"),
+                    self.tr(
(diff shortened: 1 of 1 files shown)


marmarek:
I'm going to run this through openQA just to see if nothing else breaks. We do not have tests for this specific feature.

marmarek:
Oh, we do have a test for that. And since I haven't included the [core-admin-client PR](https://github.com/QubesOS/qubes-core-admin-client/pull/161) in the test (because it fails to build), the test now hangs on exactly the message it introduces: https://openqa.qubes-os.org/tests/12043#step/system_tests/15

marmarta:
> Oh, we do have a test for that. And since I haven't included the [core-admin-client PR](https://github.com/QubesOS/qubes-core-admin-client/pull/161) in the test (because it fails to build), the test now hangs on exactly the message it introduces: https://openqa.qubes-os.org/tests/12043#step/system_tests/15

I've just fixed the core-admin-client, so it should be ready for next pass.

qubesmanager/qube_manager.py
```diff
@@ -1318,7 +1318,17 @@ def action_open_console_triggered(self):
     def action_set_keyboard_layout_triggered(self):
         # pylint: disable=invalid-name
         for vm_info in self.get_selected_vms():
-            vm_info.vm.run('qubes-change-keyboard-layout')
+            if vm_info.vm.features.check_with_template(
+                    "supported-service.keyboard-layout"):
```
marmarek, line 1322: This is now `supported-feature.keyboard-layout`. But also, add the second argument to check_with_template (default value), otherwise you'll get an exception if it is absent.

marmarek (changes requested):

marmarek (approved):

PR #273, opened 2020-11-07
Template and network change with context menu

donob4n:
Hi,

I think it's better to discuss here since https://github.com/QubesOS/qubes-issues/issues/5679 is very long already.

Checkboxes seems a nice visual and also informs you about the current template for this VM:
menu1

If you have more VM's selected it shows all the templates contained in the selection:
menu2

(of course if you select one it will change all of them to the same)

What do you think? If you like I can add the same for set Network vm.


donob4n:
Well I added a 'Network' menu too. If you try to set a halted netvm to a running VM it asks the user if wants to start the netvm.

ninavizz:
Checkboxes as a form element are not common in context menus. A simple "Check" next to the item that is the selected version, conversely, is—and I'd recommend that. Can do a mockup later today.

ninavizz:
Also, why would 2 items be selected—if the goal is to change a VM's Template from one to another?

donob4n:
> Checkboxes as a form element are not common in context menus. A simple "Check" next to the item that is the selected version, conversely, is—and I'd recommend that. Can do a mockup later today.

Ok I will try to change it.

> Also, why would 2 items be selected—if the goal is to change a VM's Template from one to another?

The idea is that it shows the template that currenty uses the selected VM, but you can select many VM's at the same time (for doing the same operation on all of them) so I don't see many options in that case. The only another option would be don't show anyone as checked.

ninavizz:
> The idea is that it shows the template that currenty uses the selected VM, but you can select many VM's at the same time (for doing the same operation on all of them) so I don't see many options in that case. The only another option would be don't show anyone as checked.

Got it! Ok, in this case, then show the multiple Template currently used by the multi-selected qubes as a bullet, and when you make a new selection, it should be a checkmark. Will the user get a dialog to confirm their choice, as this feels like a bit if a big deal change? Honestly, I think a dialog would be good for any change of a qube's designated Template VM.... unless this is something people do often and that'd be pesky?

donob4n:
> Got it! Ok, in this case, then show the multiple Template currently used by the multi-selected qubes as a bullet, and when you make a new selection, it should be a checkmark. Will the user get a dialog to confirm their choice, as this feels like a bit if a big deal change? Honestly, I think a dialog would be good for any change of a qube's designated Template VM.... unless this is something people do often and that'd be pesky?

It is not very common but there is a periodic case when your most used template reaches EOL and you have many VM's to switch to a newer version. I am agree with the dialog confirmation, also for changing network vm?

donob4n:
I don't see any way to directly use some alternative to checkboxes on the menu with the qt framework. Also It doesn't seems so uncommon, see gnome-terminal e.g.:
gnome-terminal

donob4n:
> Got it! Ok, in this case, then show the multiple Template currently used by the multi-selected qubes as a bullet, 

Do you mean something like this?
multi

ninavizz:
This is the pattern I'm speaking to, wrt checkmarks and bullets. That said, this is a pattern that has a mouseover state, a single-selected state, and a multi-selected state. The bullets indicate multiple things for the multiple items. The checkmark indicates a 1:1 correlation—hence the lighter checkmark on selection (or no checkmark at all on selected-state).

CheckboxesBullets

The square outlines indicating an actionable UI element "thing" I find to be fussy and not necessary... but if that is a standard pattern across the Qubes UI, then I'd prefer to keep them for now. 

Gnome's UI is not a "common" one for user mental models. It is among Linux users, but a goal of mine with Qubes is to make all the Qubes OS UI more consistent with common UX patterns so user migration from Windows/Mac environments is easier, while Linux users still easily "get it." Terminal is also an app required to use for how Qubes is today, but its paradigms are not standard/universal usability paradigms. Again, those goals of mine may far exceed this PR, and I'd totally respect that if so.

marmarta:
Well, Windows has a bunch of checkboxes in menus (even in the most basic of all apps, notepad) :) Given how few of our users come from Max background, I think we can assume that if it's a known Windows and Gnome pattern it's known for people. 

Not my screenshot, but a good example: 
image

ninavizz:
It's the solid-square and no disambiguation between checkmarks and bullets, that I'm more concerned with from a usability POV. The "boxes" themselves is more of an aesthetic preference. :D

donob4n:
Well, I can use setIcon(). I tested with green icon (same that marks running VM's) for single selection and yellow (transient) for multiple. Do you want to provide some differents for use on this?

These are the confirmation dialogs:
confirm2
confirm1

Do you prefer something different?

marmarek:
Offtopic: is that KDE? This blue "Yes" button is what @ninavizz asked about in https://github.com/QubesOS/qubes-issues/issues/6167

donob4n:
Yes, it's KDE :smile:

qubesmanager/qube_manager.py
```diff
@@ -794,6 +799,57 @@ def __init__(self, qt_app, qubes_app, dispatcher, _parent=None):
...
+        reply = QMessageBox.question(
+            self, self.tr("Template Change Confirmation"),
+            self.tr("Do you want to change '{0}'<br>"
+                "to Template <b>'{1}'</b>?").format(
+                ', '.join(vm.name for vm in selected_vms), template),
+            QMessageBox.Yes | QMessageBox.Cancel)
+
+        if reply == QMessageBox.Yes:
+            for info in selected_vms:
+                info.vm.template = template
```
marmarek, line 813: Setting template may fail (for example when VM is running, or when there is some issue with the template) - you need to handle errors here. The current state is the first error will interrupt the operation and leave other VMs unchanged. This may be ok if properly reported, but another option would be to collect errors for each VM individually and display summary at the end (if applicable).

qubesmanager/qube_manager.py
```diff
@@ -794,6 +799,57 @@ def __init__(self, qt_app, qubes_app, dispatcher, _parent=None):
...
+                    QMessageBox.Yes | QMessageBox.Cancel)
+
+                if reply == QMessageBox.Yes:
+                    with common_threads.busy_cursor():
+                        netvm.vm.start()
+                else:
+                    return
+
+            for info in self.get_selected_vms():
+                info.vm.netvm = netvm_name
```
marmarek, line 846: Same comment about error handling as for template.

marmarek (changes requested):

qubesmanager/qube_manager.py
```diff
@@ -807,9 +810,88 @@ def __init__(self, qt_app, qubes_app, dispatcher, _parent=None):
...
+                            netvm.vm.start()
+                    else:
+                        return
+
+            failed = []
+            for info in self.get_selected_vms():
+                try:
+                    info.vm.netvm = netvm_name
+                except:
+                    failed.append(info.name)
```
marmarek, line 874: Can you also save error message?

qubesmanager/qube_manager.py
```diff
@@ -807,9 +810,88 @@ def __init__(self, qt_app, qubes_app, dispatcher, _parent=None):
...
+                ', '.join(vm.name for vm in selected_vms), template),
+            QMessageBox.Yes | QMessageBox.Cancel)
+
+        if reply == QMessageBox.Yes:
+            failed = []
+            for info in selected_vms:
+                try:
+                    info.vm.template = template
+                except:
+                    failed.append(info.name)
```
marmarek, line 828: Here too.

marmarek (changes requested):

qubesmanager/qube_manager.py
donob4n: Yes, no problem.

donob4n: Now it does a new dialog per error. Maybe better if errors messages are too long, maybe worse if a lot of VM fail.

In any case I guess this will pretty rare to fail (if VM is started it does not let you to change the template) and when happens the error will be caught.

donob4n:
Oohh, I didn't see the conflicts :rofl:

marmarek:
PipelineRetry

qubesmanager/qube_manager.py
```diff
@@ -806,9 +809,74 @@ def __init__(self, qt_app, qubes_app, dispatcher, _parent=None):
...
+
+        errors = []
+        for info in self.get_selected_vms():
+            try:
+                info.vm.netvm = netvm_name
+            except exc.QubesValueError as ex:
+                errors.append((info.name, str(ex)))
+
+        for error in errors:
+            QMessageBox.warning(self, "{0} network change failed!"
```
marmarek, line 871: Wrap the message in `self.tr()`, to allow it be translated.

marmarek (changes requested):

marmarek (approved):

marmarek:
Some manual test results:
 - :heavy_check_mark: template change works fine for AppVMs; is inactive for DispVM (which is fine for now, since that was fixed very recently)
 - :x: netvm change doesn't support "default" option - neither for displaying current value nor for changing it to "default"

donob4n:
@marmarek How can I get current default netvm?

qubes_app.property_get_default('default_netvm') returns None

marmarek:
> qubes_app.property_get_default('default_netvm') returns None

You do that on a VM: `vm.property_get_default('netvm')` (technically it is possible that different VMs have different default, but currently it doesn't happen)

qubesmanager/qube_manager.py
```diff
@@ -869,6 +940,33 @@ def fill_cache(self):
...
+                action.setData(vm.name)
+                action.triggered.connect(partial(self.change_template, vm.name))
+
+    def _get_default_netvm(self):
+        for vm in self.qubes_app.domains:
+            if vm.klass == 'AppVM':
+                return vm.property_get_default('netvm')
+
+    def init_network_menu(self):
+        default = self._get_default_netvm().name
```
marmarek, line 957: this may fail - default netvm can be none

qubesmanager/qube_manager.py
```diff
@@ -869,6 +940,33 @@ def fill_cache(self):
...
+        self.network_menu.clear()
+        action = self.network_menu.addAction("None")
+        action.triggered.connect(partial(self.change_network, None))
+        action = self.network_menu.addAction("default ({0})".format(default))
+        action.triggered.connect(partial(self.change_network, 'default'))
```
marmarek, line 962: I'd recommend using `qubesadmin.DEFAULT` value instead of string `'default'`. This way you won't confuse it with VM named "default" (if it would be legal), but also you can normally assign it, instead of special handling.

marmarek (changes requested):

qubesmanager/qube_manager.py
donob4n: I tried using it but I can't show the network name in the confirmation dialog.

marmarek (changes requested):
Displaying default netvm doesn't work - right click on a qube that has default netvm results in network submenu that has nothing selected

marmarek:
Setting default works.

donob4n:
> Displaying default netvm doesn't work - right click on a qube that has default netvm results in network submenu that has nothing selected

Fixed

donob4n:
Maybe it would be a good idea to disable network menu for templates, even enforce a warning in 'Settings' warning that setting a network will break a base pillar of Qubes security.

marmarek (approved):

qubesmanager/qube_manager.py
```diff
@@ -746,7 +750,6 @@ def __init__(self, qt_app, qubes_app, dispatcher, _parent=None):
             column = self.qubes_model.columns_indices[col_no]
             action = self.menu_view.addAction(column)
             action.setData(column)
-            action.setCheckable(True)
```
marmarek, line 749: This one should stay. I'll fix that when merging.

donob4n: Ohh yes, I noticed that but forgot to fix great :)

PR #275, opened 2020-11-14
Use checkbox for 'Include in Backup' column

donob4n:
It was really simple thanks to model/view architecture.

I tried to center the checkbox on the column but it will require to subclass ItemDelegate or similar. Maybe it would be just fine with a short column header text (e.g. "Inc. in Backup") or using two lines instead one.

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -405,8 +405,6 @@ class QubesTableModel(QAbstractTableModel):
                 return "Yes" if vm.internal else ""
             if col_name == "IP":
                 return vm.ip
-            if col_name == "Include in backups":
-                return "Yes" if vm.inc_backup else ""
             if col_name == "Last backup":
                 return vm.last_backup
             if col_name == "Default DispVM":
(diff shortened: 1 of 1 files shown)


marmarek:
> "Inc. in Backup"

Maybe just "Backup"? Possibly with a tooltip "Include in Backup"?

qubesmanager/qube_manager.py
```diff
@@ -477,6 +472,27 @@ def headerData(self, col, orientation, role):
...
 
+    def setData(self, index, value, role=Qt.EditRole):
+        if not index.isValid():
+            return False
+
+        if role == Qt.CheckStateRole:
+            col_name = self.columns_indices[index.column()]
+            if col_name == "Include in backups":
+                vm = self.qubes_cache.get_vm(index.row())
+                vm.vm.include_in_backups = (value == 2)
```
marmarek, line 483: What is value `2`? Isn't there some nice constant like `Qt.Checked`?

marmarek (changes requested):

donob4n:
> > "Inc. in Backup"
> 
> Maybe just "Backup"? Possibly with a tooltip "Include in Backup"?

With just 'Backup' looks nice. I'm not sure about what to do with the tooltip. No other headers or cells have tooltips so it can be hard to found. In case you prefer to add it, where do you see more useful in the column header or in all 'Backup' cells?

marmarek:
PipelineRetry

marmarek:
PipelineRetry

marmarek (approved):

PR #277, opened 2020-12-05
Multiple file log dialog

donob4n:
Not very fashion but it saves a lot of clicks if you want to see many log files from one (or more) qubes.

logdialog

Probably a list should be a better solution than buttons.

### DIFF
--- a/qubesmanager/log_dialog.py
+++ b/qubesmanager/log_dialog.py
@@ -58,7 +58,7 @@ class LogDialog(ui_logdlg.Ui_LogDialog, QtWidgets.QDialog):
                 button = QtWidgets.QPushButton(log_path)
                 button.clicked.connect(partial(self.set_current_log, log_path))
                 self.buttonsLayout.addWidget(button,
-                        count / butts_in_row, count % butts_in_row)
+                        count / btns_in_row, count % btns_in_row)
                 count += 1
 
         if count == 0:
--- a/qubesmanager/log_dialog.py
+++ b/qubesmanager/log_dialog.py
@@ -20,13 +20,12 @@
 #
 #
 import sys
+import os
+from functools import partial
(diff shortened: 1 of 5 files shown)


qubesmanager/log_dialog.py
```diff
@@ -52,8 +51,32 @@ def __init__(self, app, log_path, parent=None):
         self.__init_log_text__()
 
     def __init_log_text__(self):
+        butts_in_row = 3
```
marmarek, line 54: :smile: 
```suggestion
        btns_in_row = 3
```

marmarek (changes requested):

qubesmanager/log_dialog.py
donob4n: lol

marmarek: But you need to update usages of the variable too.

donob4n: Yes sorry, it seems that there are also conflicts with other PR. I will check tonight.

marmarek:
If there are no logs, the logs window is still opened after a message.

donob4n:
Yes I noticed  it when I added the warning dialog. I tried to close the window using accept(), close(), reject(),... but all failed probably because they were called from _init__(). Maybe the best solution is moving the file checking to qube_manager.py and show the error from there.

marmarek:
On Thu, Dec 10, 2020 at 03:06:32PM -0800, donob4n wrote:
> Yes I noticed  it when I added the warning dialog. I tried to close the window using accept(), close(), reject(),... but all failed probably because they were called from _init__(). Maybe the best solution is moving the file checking to qube_manager.py and show the error from there.

Yes.

marmarek (approved):

PR #278, opened 2020-12-09
Added template and dispvm to create dialog

ctrlaltf24:
Adds TemplateVM and DisposableVMs to the Create New Qube dialog.
Changes:
- Rewording of VM class dropdown to make more sense in context
- Added none to template rather than having a separate type for empty
- Adds TemplateVM and DisposableVM
- Adds dynamic updating of the template list depending on the selected VM class.

Considerations before merging:
- Is the use of in acceptable? (line 194)
- Is the dynamic updating of the template dropdown a reasonable solution?
- Does anything need to be updated on the translation side?

### DIFF
--- a/qubesmanager/create_new_vm.py
+++ b/qubesmanager/create_new_vm.py
@@ -59,7 +59,7 @@ class CreateVMThread(QtCore.QThread):
 
                 vm = self.app.clone_vm(self.template, self.name,
                                        self.vmclass, **args)
-                
+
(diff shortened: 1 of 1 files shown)


marmarek:
Pylint (correctly) complains about trailing spaces.

marmarek:
PipelineRetry

marmarek:
PipelineRetry

marmarek (approved):

PR #279, opened 2020-12-24
Fix build with python3 for CentOS

fepitre:


marmarek:
PipelineRetry

marmarek:
PipelineRetry

marmarek:
PipelineRetry

PR #281, opened 2021-01-10
Cascade shutdown

donob4n:
Fix: https://github.com/QubesOS/qubes-issues/issues/6109

I'm using time.sleep() which is not ideal. 
Is there some better way for doing a sync vm.shutdown()? If not, would it be nice to add it?


donob4n:
Ouch! It can cause a infinite loop.

marmarek:
> Is there some better way for doing a sync vm.shutdown()?

The actual function on the qubesd side do support waiting for shutdown to complete. It shouldn't be hard to expose it in qubesadmin module.

marmarek:
That said, it may be better to first initiate shutdown of all connected VMs and only then wait for all of them to actually shutdown, instead of doing it one by one.

donob4n:
> > Is there some better way for doing a sync vm.shutdown()?
> 
> The actual function on the qubesd side do support waiting for shutdown to complete. It shouldn't be hard to expose it in qubesadmin module. I'll open a PR.
Do you think that current master is compatible with R4.0? If yes I could try to do it.

marmarek:
As is - unlikely. But you can try, it's possible that this area may work with some minor changes.

donob4n:
I've tried both adding a 'Cascade' and 'Wait' arguments but it seems pretty complex without a full understanding on the different. "qvm_shutdown" seems manually waiting for shutdown instead blocking on 'vm.shutdown(force=force)' call.

And I'm also running R4.0, better I will let this version for now and try to fix in a better way when I finally upgrade to R4.1

donob4n:
This way it can cause an infinite loop if some VM fails to shutdown. I've tried to replace the 'time.sleep()' with 'asyncio.wait_for(events_utils.wait_for_domain_shutdown(),time)' but I'm doing something wrong with asyncio/pyqt coexistence.

I will try to fix it.

donob4n:
Well, after failing with async/await/run_until_zzz/... I found a better solution, don't wait! just force shutdown.

If the user wants a cascade shutdown I think it will not cause problems.

qubesmanager/qube_manager.py
```diff
@@ -1249,13 +1249,33 @@ def action_shutdownvm_triggered(self):
...
+            connected_vms = [x for x in vm.connected_vms if x.is_running()]
+
+            if len(connected_vms) > 0:
+                reply = QMessageBox.question(
+                    self, self.tr("Qube Shutdown Confirmation"),
+                    self.tr("There are some qubes connected to <b>'{0}'</b>!"
+                        "<br><small>Do you want to shutdown: </small>"
+                        "<b>'{1}'</b>?").format(vm.name,
+                            ", ".join([x.name for x in connected_vms])),
+                    QMessageBox.Yes | QMessageBox.Cancel)
```
marmarek, line 1261: If you for example try to shutdown sys-net, it will first ask this question with sys-firewall as a connected VM, and if you confirm, it will ask again, with other VMs this time. I think it should rather list all of them in the first confirmation.
On the other hand, if you want to just restart a network-providing VM (like sys-net or sys-firewall), it should work in most cases without the need to shutdown connected VMs (but you need to pass force=True).

marmarek (changes requested):

qubesmanager/qube_manager.py
donob4n: Should we ask for additional confirmation if the VM has connected VMs? (in restart case)

donob4n: Since restarting is not very dangerous (for the connected vms) I think that asking again is not necessary.

donob4n:
I added small commit for avoid vm's disappear while filtering 'running' and 'halted'.

qubesmanager/qube_manager.py
```diff
@@ -1246,16 +1246,44 @@ def action_shutdownvm_triggered(self):
...
+
+                if reply != QMessageBox.Yes:
+                    return False
+
+                force = True
+                shutdown_time = shutdown_time * len(connected_vms)
```
marmarek, line 1276: This looks a bit weird when shutting down all the VMs in parallel. If for example you have 10 of them, that shouldn't take 10x longer. The default shutdown timeout (60s) is already quite liberal, so maybe use just that (you are risking just a prompt to the user, not automatic kill), or max from all the involved VMs? If that really isn't enough, then maybe x2?

donob4n: uhM.. the default is 20s. I noticed the warning appearing with only 2 simultaneous shutdowns. Maybe we could multiply the total affected vms but set a max limit. maybe 1min?

marmarek: Oh, indeed qubes manager does not use `shutdown_timeout` VM property, but its own hardcoded value of 20s... Maybe change that?

donob4n: lol

Is that enough? lol I don't know what happened :D

marmarek: I think qt uses ms not s for timers

marmarek:
Please update tests too (see https://openqa.qubes-os.org/tests/16501/file/system_tests-tests-qubesmanager.tests.log)

donob4n:
Ok I will try to fix this night.

donob4n:
I've fixed all tests except:
- test_204_vm_keyboard
It seems that `vm_info.vm.features.check_with_template("supported-feature.keyboard-layout", False)` fails although `qubes-change-keyboard-layout` works properly, any idea?

- test_500_logs
Well this compared the `logs_menu.actions()` between dom0 and another AppVM, but currently this data is in a local variable inside `action_show_logs()` and a member of LogDialog(). I'm gonna try to fix it catching the new LogDialog() and accessing to `self.logfiles` but I'm newbie with Mocks and this test engine so maybe I end with some simple workaround (adding logfiles as vm_info member?)

donob4n:
Ouch, I will fix the conflicts. I saw that probably the keyboard problem is already solved, maybe also the show_log tests was fixed in another way.

marmarek:
This is the test result of v4.1.15: https://openqa.qubes-os.org/tests/16562 - all passes there.

donob4n:
I see that you fixed the tests broken by the other PR's and created newer ones. I re-fixed some of them but at least I learnt something about this crazy Mocks :) 

I enabled 'All Activity' notifications in the hope it notices any commit.

qubesmanager/qube_manager.py
donob4n: @marmarek
> On the other hand, if you want to just restart a network-providing VM (like sys-net or sys-firewall), it should work in most cases without the need to shutdown connected VMs (but you need to pass force=True).

One Windows HVM crashed after restarting its network VM. Maybe we should test a little this before merging. I will check with Linux HVM's and version 4.1 instead 4.0.

donob4n: I tested with two Linux based (Alpinelinux and Fedora) and both worked perfect when rebooting sys-firewall. I can't check with Windows guest since I can't boot a restored Windows VM.

donob4n: Well, at least it doesn't crash but network doesn't work with Windows 7.  Maybe should it show a Warning and ask for confirmation? Do you think it could be fixed?

donob4n:
Is there some problem with this?

marmarek:
Please update the test too: https://openqa.qubes-os.org/tests/17437#step/QubeManagerTest/31

donob4n:
Ey @marmarek I think that the test is pretty unstable because it will change if the selected vm has connected vm's or not. So I think that forcing '_select_non_admin_vm()'  to choose one which doesn't provide network should fix this. See https://github.com/donob4n/qubes-manager/blob/4efc46de64ad0e11cf2778746e5d9dc896b8b134/qubesmanager/tests/test_qube_manager.py#L1485

Unfortunately I migrated my desktop to Alpinelinux so I can't check if it works and doesn't break other things.

marmarek:
That should indeed work too, thanks. I'll re-run the test shortly (but it takes some hours to complete).

marmarek:
After making sure there actually is some non-network-providing VM running, it works :)

marmarek:
PipelineRetry

marmarek (approved):

PR #287, opened 2021-02-27
Repro fc

fepitre:


marmarek:
I see vm-fc33 is reproducible but dom0-fc32 is not. Do you know why and is it realistic to make dom0-fc32 reproducible too? If not, maybe drop it from the CI config?

fepitre:
Fixed :)

marmarek (approved):

PR #288, opened 2021-03-15
VMBootFromDeviceWindow improvement and other minor fixes

donob4n:
- VMBootFromDeviceWindow() doesn't start the VM itself. It now "returns" the cdrom_location so caller can handle it in a better way:
- Improved dialog size and margins
- The dialog appears now very fast.
- If the user decides to cancel it, the VM is not created. It returns to the previous dialog so he can modify something and continue, or cancel the whole process creation.

Apart from this I added a properly **parent** argument to all dialogs (except current 'Template Manager', is it going to be removed?). This is very welcome when working with big screens. Now all dialogs are created centered on the current Qube Manager window.

### DIFF
--- a/qubesmanager/about.py
+++ b/qubesmanager/about.py
@@ -30,8 +30,8 @@ from . import ui_about  # pylint: disable=no-name-in-module
 
 # pylint: disable=too-few-public-methods
 class AboutDialog(ui_about.Ui_AboutDialog, QDialog):
-    def __init__(self):
-        super().__init__()
+    def __init__(self, parent=None):
+        super().__init__(parent)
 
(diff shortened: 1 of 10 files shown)


marmarek (changes requested):
You need to adjust also `main()` function, otherwise standalone call `qubes-vm-boot-from-device <vmname>` doesn't work.

donob4n:
There is a:
`qubesmanager/bootfromdevice.py:172:0: C0301: Line too long (81/80) (line-too-long)`
but I feel bad restarting all the pipeline :\

marmarek (approved):

PR #292, opened 2021-05-10
Avoid spurious warnings when setting NetVM

SaswatPadhi:
Resolves QubesOS/qubes-issues#6592.

Qube Manager currently shows a security warning when changing the NetVM of TemplateVMs.
This warning is generally useful in other cases, but not when disabling networking;
for example changing from from "default" to "none".

In this commit, I introduce a change to not warn when disabling networking, by setting NetVM to None.

### DIFF
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -627,7 +627,14 @@ class VMSettingsWindow(ui_settingsdlg.Ui_SettingsDialog, QtWidgets.QDialog):
             self.warn_too_much_mem_label.setVisible(True)
 
     def check_warn_templatenetvm(self):
-        if self.vm.klass == 'TemplateVM':
+        if self.vm.klass != 'TemplateVM':
+            return
+
+        current_netvm = self.netVM.currentData()
+        if current_netvm is None:
+            return
+
+        if current_netvm != qubesadmin.DEFAULT:
             QtWidgets.QMessageBox.warning(
                 self,
                 self.tr("Warning!"),


qubesmanager/settings.py
```diff
@@ -625,17 +625,25 @@ def check_mem_changes(self):
...
+        if current_netvm != qubesadmin.DEFAULT or default_netvm:
             QtWidgets.QMessageBox.warning(
                 self,
                 self.tr("Warning!"),
-                self.tr(
-                    "Connecting a TemplateVM directly to a network is highly"
-                    " discouraged! <br> <small>You are breaking a basic part "
-                    "of Qubes security and there is probably no real need"
-                    " to do so. Continue at your own risk.</small>"))
+                self.tr("Connecting a TemplateVM or DVM Template directly to a "
```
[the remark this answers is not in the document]
SaswatPadhi: Ah yes, makes sense! I was thinking of some of the static DispVMs I have.

DemiMarie:
Minor nit: for @QubesOS repos, we generally prefer that fixup commits be squashed into the commit they fix.

SaswatPadhi:
Oh, I kept the commits separate so the changes after each review pass are easy to see. GitHub allows squashing all commits in the PR before merging.

DemiMarie:
> Oh, I kept the commits separate so the changes after each review pass are easy to distinguish. GitHub allows squashing all commits in the PR before merging.

We don’t use GitHub’s squash-merge feature, as it invalidates commit signatures.  All of our merges are done client-side.

SaswatPadhi:
Squashed all the commits.

@DemiMarie Please take another look.

SaswatPadhi:
May be @marmarek could take a look as well?

Thank you.

qubesmanager/settings.py
```diff
@@ -627,7 +627,15 @@ def check_mem_changes(self):
...
-        if self.vm.klass == 'TemplateVM':
+        if self.vm.klass != 'TemplateVM':
+            return
+
+        current_netvm = self.netVM.currentData()
+        if current_netvm is None:
+            return
+
+        default_netvm = self.vm.property_get_default('netvm')
+        if current_netvm != qubesadmin.DEFAULT or default_netvm:
```
marmarek, line 638: Did you meant `current_netvm != default_netvm`?

marmarek (changes requested):
Also, please update the commit message - it doesn't do anything about DVM Templates.

qubesmanager/settings.py
SaswatPadhi: I am trying to check:

Either `current_netvm` is _not_ set to `(default)` on the Qube Manager, or (if it is then) `default_netvm` is not `None`

If this check succeeds, then the VM has networking enabled.

If you have any suggestions on making the check cleaner, I'd be happy to change it.

SaswatPadhi: Simply checking `current_netvm != default_netvm` wouldn't work in all cases.

Consider the case when `default_netvm` is not set to `None` in global settings, say it's set to `vm-1`. Then the check wouldn't trigger when we set a template's NetVM to `vm-1` although the template would indeed have networking enabled.

DemiMarie: The default is always for TemplateVMs to have no network, regardless of what `default_netvm` is set to.

SaswatPadhi: Yes, the default NetVM for TemplateVMs is set to `(none)` in the Qube Manager. But that's different from setting it to `(default)`, which is the `default_netvm`.

In this check, I essentially avoid a spurious warning if a user changes the NetVM to `(default)` and `default_netvm` is set to `(none)` in global settings.

Do you think this is reasonable?

DemiMarie: For TemplateVMs, `(none)` and `(default)` *should* be equivalent, regardless of what `default_netvm` is.

SaswatPadhi: I'm afraid that they don't seem to be equivalent.

Here is what I see on my system:
`(default)` takes the value set in the global settings, and shows up as `default (vm-1)` for example, if `vm-1` is the `default_netvm` from global settings.

SaswatPadhi: I actually don't think they should be equivalent, because that would mean `(default)` has a different meaning for TemplateVMs (always `(none)`), than AppVM (`default_netvm`).

DemiMarie: > I actually don't think they should be equivalent, because that would mean `(default)` has a different meaning for TemplateVMs (always `(none)`), than AppVM (`default_netvm`).

It actually does have a different meaning, precisely because setting a NetVM for a TemplateVM is almost always a bad idea.

SaswatPadhi: Right, I see your point.

I guess the question here is, currently `(none)` and `(default)` are not equivalent. Should we make them equivalent?

I am happy to remove the check from my PR if we decide to do so.

DemiMarie: Internally, `(none)` maps to `netvm` being set to the empty string (which means “no NetVM”), while `(default)` means that `netvm` is not set at all.  An unset property is interpreted as having a default value.  For TemplateVMs, the default is networking disabled, while for other types of VMs, the default is set by the `default_netvm` global setting.

In short, both `(none)` and `(default)` are okay.  Anything else should give a warning.

SaswatPadhi: Thanks.

So I should simplify the check to just `if current_netvm != qubesadmin.DEFAULT:`, right?

DemiMarie: > Thanks.
> 
> So I should simplify the check to just `if current_netvm != qubesadmin.DEFAULT:`, right?

Exactly

SaswatPadhi: Fixed as suggested.

qubesmanager/settings.py
```diff
@@ -627,7 +627,15 @@ def check_mem_changes(self):
...
     def check_warn_templatenetvm(self):
-        if self.vm.klass == 'TemplateVM':
+        if self.vm.klass != 'TemplateVM':
+            return
+
+        current_netvm = self.netVM.currentData()
+        if current_netvm is None:
+            return
+
+        default_netvm = self.vm.property_get_default('netvm')
```
DemiMarie, line 637:
```suggestion
```

DemiMarie (changes requested):
The `default_netvm` line is now dead code 🙂.

qubesmanager/settings.py
SaswatPadhi: Ah yes!! Good catch. Thanks 😄

DemiMarie: You’re welcome!  Thanks for the contribution.

SaswatPadhi:
I've updated the commit message and addressed all code review comments.
Please take another look, @marmarek.

DemiMarie (approved):
Looks good!

marmarek (approved):

marmarek:
PipelineRetry

PR #300, opened 2021-09-08
Improved design of Global Settings

marmarta:
Transitional form: the functionality is the same, changes are
mostly visual, with only a minor addition of Version no,
release info link and version link.

fixes QubesOS/qubes-issues#6858

### DIFF
--- a/debian/install
+++ b/debian/install
@@ -40,6 +40,8 @@
 
 /usr/lib/*/dist-packages/qubesmanager/resources_rc.py
 
+/usr/lib/*/dist-packages/qubesmanager/global_settings.css
+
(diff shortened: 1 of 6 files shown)


marmarek:
Font sizes here are huge, it barely fits on the screen (and I got a scrollbar). Can you use the same font size as in the manager? I've just checked - it _does_ use whatever you set in global appearance settings, so probably it will be enough to simply not override it with any explicit value.

And also - I don't like the new buttons at the bottom, because:
 - they are in different place than in every other dialog
 - they have _only_ text and no graphical (or color) indicator of the action - again contrary to every other dialog in the system

PR #303, opened 2021-10-06
Add '?' to allowed characters in file paths in Manager tools

marmarta:
Also made the boot-from-device window resizeable
in a more functional way.

fixes QubesOS/qubes-issues#6438

### DIFF
--- a/qubesmanager/utils.py
+++ b/qubesmanager/utils.py
@@ -420,7 +420,7 @@ def get_path_from_vm(vm, service_name):
     :return: path to file, checked for validity
     """
 
-    path_re = re.compile(r"[a-zA-Z0-9/:.,_+=() -]*")
+    path_re = re.compile(r"[a-zA-Z0-9/:.,_+=() ?-]*")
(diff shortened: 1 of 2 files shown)


marmarta:
I have added question mark as an additional character, but at the moment many other characters will also cause failure. It would be perhaps best to rethink our entire path-checking function:
```
    path_re = re.compile(r"[a-zA-Z0-9/:.,_+=() -?]*")
    path_max_len = 512

    if not vm:
        return None
    stdout, _stderr = vm.run_service_for_stdio(service_name)

    stdout = stdout.strip()

    untrusted_path = stdout.decode(encoding='ascii')[:path_max_len]

    if not untrusted_path:
        return None
    if path_re.fullmatch(untrusted_path):
        assert '../' not in untrusted_path
        assert '\0' not in untrusted_path
        return untrusted_path.strip()
    raise ValueError(QtCore.QCoreApplication.translate(
        "ManagerUtils", 'Unexpected characters in path.'))

```

PR #310, opened 2022-02-20
Update qube_manager.py

logoerthiner1:
Catch subprocess.CalledProcessError rather than let the error abort.

It should fix the issue when qube-manager aborts when executed command exits with error.

The issue is mentioned many times in many places. See https://github.com/QubesOS/qubes-issues/issues/7278

DISCLAIMER: not tested, as I do not have a good testing environment


marmarek:
pylint complains:
```
************* Module qubesmanager.qube_manager
qubesmanager/qube_manager.py:630:0: C0301: Line too long (92/80) (line-too-long)
qubesmanager/qube_manager.py:643:0: C0301: Line too long (92/80) (line-too-long)
```

marmarek:
PipelineRetry

marmarek (approved):

PR #316, opened 2022-04-02
Fix issues found by pylint

marmarek:
Use context manager for handling files and some processes.
Better iterate over dicts and arrays.


marmarek:
PipelineRetry

PR #320, opened 2022-05-05
Refresh VM state at connection established in manager

marmarta:
fixes QubesOS/qubes-issues#7362


DemiMarie:
@marmarta looks like you need to add some pylint suppressions and wrap a long line

marmarek (approved):

PR #324, opened 2022-09-28
Fix VM sorting order in Qube Manager if Sorting by State and fix log browser

marmarta:
Previously the sorting order depended on alphabetically sorting the dictionary containing state, which was less than ideal. Now it will use a reasonable order (from more to less running, and within a given state, no updates come before updates-available).
Also fixes sorting the Backup column to actually use the include_in_backups state and the Label column to sort by label and type.

Also fixes the log browser to be selection-aware.

fixes QubesOS/qubes-issues#7073
fixes https://github.com/QubesOS/qubes-issues/issues/7633
fixes https://github.com/QubesOS/qubes-issues/issues/7776
fixes https://github.com/QubesOS/qubes-issues/issues/7715

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -467,7 +467,18 @@ class QubesTableModel(QAbstractTableModel):
             if col_name == "Label":
                 return str(vm.label)
             if col_name == "State":
-                return str(vm.state)
+                # sorting order is based on a logical order (from running to
+                # progressively less running) and update state
+                state = vm.state.get('power', '')
+                try:
+                    ordered_state = str(
+                        ["Running", "Transient", "Halting", "Paused",
+                         "Suspended", "Dying", "Crashed",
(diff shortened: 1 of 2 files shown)


qubesmanager/qube_manager.py
```diff
@@ -465,11 +465,32 @@ def data(self, index, role):
             if col_name == "Type":
                 return vm.klass
             if col_name == "Label":
-                return str(vm.label)
+                vmtype, vmcolor = vm.icon.split("-")
+                try:
+                    processed_color= str([
+                        "red", "orange", "yellow", "green", "gray", "blue",
+                        "purple", "black"].index(vmcolor))
```
marmarek, line 472: `vm.label` is a `Label` object, which has `index` property. Better use that.

marmarek (changes requested):

marmarta:
fixed

marmarek (approved):

PR #326, opened 2022-11-11
Allow saving backup settings without passphrase

meithecatte:
This adds a separate "save passphrase in backup profile" checkbox, which is unchecked by default and gets saved in the profile (implicitly, with the presence of the password field). As a result, users may save their settings without having to compromise on the security of their passphrase.

The big fat warning about the passphrase getting saved like that is now grayed out when not applicable.

Fixes QubesOS/qubes-issues#4819

image


marmarek:
The backup profile is how backup code gets the password from the user, so it needs to be written there for making the backup. But it can be written into some temporary file (tmpfs or such) and only symlinked into actual backup profile location.

meithecatte:
Yes, the actual temporary profile that gets passed to the backup tool does include the password. Improving this, I believe, is best left for a future PR.

marmarek (approved):

PR #328, opened 2022-11-21
Add explanation why you can't use vm settings with AdminVMs

marmarta:
fixes QubesOS/qubes-issues#7897

### DIFF
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -26,6 +26,7 @@ import collections
 import functools
 import re
 import subprocess
+import sys
 import traceback
 from qubesadmin.tools import QubesArgumentParser
 from qubesadmin import devices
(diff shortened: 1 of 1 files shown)


qubesmanager/settings.py
```diff
@@ -1498,6 +1498,12 @@ def delete_rule_button_pressed(self):
 def main(args=None):
     args = parser.parse_args(args)
     vm = args.domains.pop()
+    if vm.klass == 'AdminVM':
+        print("This tool cannot be used to change properties of an "
+              f"AdminVM ({vm.name}).")
+        print("You can use command-line tools such as qvm-prefs "
+              "and qvm-features to change properties of an AdminVM")
+        return 1
```
marmarek, line 1506: For this to take effect, `main()` call below needs to use the return value, like `sys.exit(main())`

marmarek (changes requested):

marmarta:
fixed

marmarek (approved):

PR #329, opened 2022-12-05
Add missing <!DOCTYPE html>

DemiMarie:
This should be at the start of all HTML documents, to avoid quirks mode.

### DIFF
--- a/ui/backupdlg.ui
+++ b/ui/backupdlg.ui
@@ -362,7 +362,7 @@
     <item>
      <widget class="QTextEdit" name="textEdit">
       <property name="html">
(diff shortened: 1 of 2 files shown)


marmarta:
Qt Designer cannot parse those !DOCTYPE directives, and this makes editing text in Qt Designer very cumbersome (cannot use Rich Text edit mode, must write HTML directly).

marmarta:
yup, probably an artifact of someone using different tools/some other random fossil. If you want to remove it with this PR, feel free!

marmarek (approved):

PR #337, opened 2023-02-03
Make available/selected apps in settings clearer

marmarta:
Replace Available/Selected with a more explicit
"Available in qube/template" and "Shown in App Menu"

fixes https://github.com/QubesOS/qubes-issues/issues/8012

based on https://github.com/QubesOS/qubes-manager/pull/335 to avoid cherrypicking the pylint fixes back and forth.

### DIFF
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -220,6 +220,9 @@ class VMSettingsWindow(ui_settingsdlg.Ui_SettingsDialog, QtWidgets.QDialog):
         ####### apps tab
         if self.tabWidget.isTabEnabled(self.tabs_indices["applications"]):
             self.app_list = multiselectwidget.MultiSelectWidget(self)
+            self.app_list.change_labels(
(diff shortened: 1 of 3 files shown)


qubesmanager/settings.py
```diff
@@ -220,6 +220,14 @@ def __init__(self, vm, init_page="basic", qapp=None, qubesapp=None,
         ####### apps tab
         if self.tabWidget.isTabEnabled(self.tabs_indices["applications"]):
             self.app_list = multiselectwidget.MultiSelectWidget(self)
+            if self.vm.template:
+                self.app_list.change_labels(
+                    available="Applications available in template",
```
marmarek, line 225: The list for template-based qubes includes apps from both its template and the app qube itself (if you have something installed in your home dir). So, a common text for both cases sounds better.

marmarek (changes requested):

marmarek (approved):

PR #339, opened 2023-02-15
Fix tests for python 3.11

marmarta:


marmarta:
pylint fails due to update, fixes in https://github.com/QubesOS/qubes-manager/pull/337

marmarek (approved):

PR #344, opened 2023-04-29
Preliminary GUI initiated QWT install

Augsch123:
Simple modification.

~~Current limitations: If there's no QWT installed, this will silently fail ( the same as previously ). However, qvm-start will leave a humble line in the journal.~~

~~I thought of using os.path.exist() to detect if there's qwt image, but I'm not sure how to correctly implement it.~~

Current limitations: Path is hardcoded, and I think it will work awkwardly in sys-gui.

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -1414,7 +1414,9 @@ class VmManagerWindow(ui_qubemanager.Ui_VmManagerWindow, QMainWindow):
                     self.tr("QWT not found"),
                     self.tr("'qubes-windows-tools' is not installed in dom0."))
         for vm_info in self.get_selected_vms():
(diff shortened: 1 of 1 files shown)


Augsch123:
This pull request is able to address [QubesOS/qubes-issues#8197](https://github.com/QubesOS/qubes-issues/issues/8197), when qubes manager is running in dom0. But when qubes manager is running in sys-gui, I'm not aware of a way to check whether the qwt image exists in dom0.

marmarek:
> This pull request is able to address [QubesOS/qubes-issues#8197](https://github.com/QubesOS/qubes-issues/issues/8197), when qubes manager is running in dom0. But when qubes manager is running in sys-gui, I'm not aware of a way to check whether the qwt image exists in dom0.

That indeed will require some smarter handling, the way `qvm-start --cdrom` is implemented won't work for ISO file at all if you call it from sys-gui. But lets merge the fix for the dom0 case first.

marmarek (approved):

PR #346, opened 2023-05-12
Check if default_netvm is running when changing netvm

marmarta:
Changing netvm checks if new netvm is running, and if not asks to start it - this didn't happen for default netvm. Now it will.

fixes QubesOS/qubes-issues#8156

### DIFF
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -562,7 +562,26 @@ class VMSettingsWindow(ui_settingsdlg.Ui_SettingsDialog, QtWidgets.QDialog):
         # vm netvm changed
         try:
             if utils.did_widget_selection_change(self.netVM):
-                self.vm.netvm = self.netVM.currentData()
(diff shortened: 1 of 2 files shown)


qubesmanager/settings.py
```diff
@@ -562,7 +562,27 @@ def __apply_basic_tab__(self):
         # vm netvm changed
         try:
             if utils.did_widget_selection_change(self.netVM):
-                self.vm.netvm = self.netVM.currentData()
+                if self.netVM.currentData() == qubesadmin.DEFAULT:
+                    netvm = self.vm.property_get_default('netvm')
+                else:
+                    netvm = self.netVM.currentData()
+                    print("netvm", type(netvm))
```
marmarek, line 569: leftover debug print

marmarek (changes requested):

qubesmanager/settings.py
marmarta: done

marmarek (approved):

PR #349, opened 2023-06-12
Fix error message when qube renaming fails.

verygreen:
Need to properly indicate the new cloned qube name

fixes QubesOS/qubes-issues#8056


DemiMarie:
PipelineRetryFailed

marmarek (approved):

PR #351, opened 2023-06-28
Improve field order in new fw window

marmarta:
fixes QubesOS/qubes-issues#8236

### DIFF
--- a/ui/newfwruledlg.ui
+++ b/ui/newfwruledlg.ui
@@ -20,6 +20,16 @@
    <bool>true</bool>
   </property>
   <layout class="QGridLayout" name="gridLayout">
+   <item row="1" column="0">
+    <widget class="QDialogButtonBox" name="buttonBox">
+     <property name="orientation">
+      <enum>Qt::Horizontal</enum>
+     </property>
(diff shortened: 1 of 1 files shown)


ui/newfwruledlg.ui
```diff
@@ -106,49 +142,12 @@
...
-     <property name="standardButtons">
-      <set>QDialogButtonBox::Cancel|QDialogButtonBox::Ok</set>
-     </property>
-    </widget>
-   </item>
   </layout>
  </widget>
  <tabstops>
   <tabstop>addressComboBox</tabstop>
-  <tabstop>serviceComboBox</tabstop>
```
marmarek, line 151: I don't think it should be dropped from here.

marmarek (changes requested):

marmarek (changes requested):
Something is still wrong with the focus. Previously, when dialog was opened, it started with focus on the address field (expected behavior), now it starts with focus on cancel (or simply nothing, and cancel happen to be default button?).

marmarta:
I hate tab order. Anyway, closing this and pushing a better one.

PR #356, opened 2023-08-03
Use subprocess.Popen to spawn global config in manager

marmarta:
Don't have the manager wait for global config to be exited.

fixes QubesOS/qubes-issues#8387


marmarek (approved):

marmarek:
for both added lines, pylint complains

PR #357, opened 2023-08-10
Update size on disk in Qube Manager occassionally

marmarta:
To avoid too big a system load, the update will happen:
- every 5 minutes for running qubes
- on qube state change (running -> transient -> halted etc.)

fixes QubesOS/qubes-issues#6053


marmarek (changes requested):
Apparently our current API does not show runtime disk usage, it only shows disk usage from before starting the VM. This is problematic, but until fixed, updating with a timer doesn't really make sense.

marmarek (approved):
Lets keep it this way, the API to report running disk usage will be fixed eventually.

PR #363, opened 2024-02-06
update to new device API

piotrbartman:
after QubesOS/qubes-core-admin/pull/579

https://github.com/QubesOS/qubes-issues/issues/4626


marmarek (changes requested):
pylint complains, otherwise looks good

marmarek (approved):

PR #365, opened 2024-06-24
Visually indicate unsupported Templates & StandaloneVMs in Qube Manager

alimirjamali:
Fixes: https://github.com/QubesOS/qubes-issues/issues/4581

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -102,6 +102,7 @@ class StateIconDelegate(QStyledItemDelegate):
                 "update" : QIcon(":/update-recommended.png"),
                 "outdated" : QIcon(":/outdated.png"),
                 "to-be-outdated" : QIcon(":/to-be-outdated.png"),
+                "eol": QIcon(':/warning.png')
                 }
         self.outdatedTooltips = {
                 "update" : self.tr("Updates pending!"),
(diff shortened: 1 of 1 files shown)


qubesmanager/qube_manager.py
```diff
@@ -239,10 +243,16 @@ def update_power_state(self):
...
-                    manager_utils.get_feature(
+            if self.vm.klass in {'TemplateVM', 'StandaloneVM'}:
+                if manager_utils.get_feature(
                         self.vm, 'updates-available', False):
-                self.state['outdated'] = 'update'
+                    self.state['outdated'] = 'update'
+                elif manager_utils.get_feature(
+                        self.vm, 'os-eol', None):
+                    d = [int(x) for x in manager_utils.get_feature(self.vm, 'os-eol', '0-0-0').split('-')]
+                    eol = datetime(d[0], d[1], d[2])
```
marmarta, line 253:
```suggestion
                        eol_string: str = vm.features.get('os-eol', '')
                        eol = datetime.strptime(eol_string, '%Y-%m-%d')
```

This is a bit more readable, but more importantly, it's the same logic as in the qui updater widget, to be sure we use the same one - that's the function https://github.com/QubesOS/qubes-desktop-linux-manager/blob/8bf07aa132dd47d1b501ec4ec2706c042f77dc7e/qui/utils.py#L88

marmarta (changes requested):

qubesmanager/qube_manager.py
alimirjamali: Applied it and tested it. With one change that it should be `self.vm.features.get`

marmarta: ah, right, you're correct

marmarta (approved):
pylint fail is due to some unrelated pylint update, I'll fix it independently

PR #367, opened 2024-06-30
Rename the Window title of old Template Manager to Template Switcher

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/9298

### DIFF
--- a/qubesmanager/i18n/qubesmanager_en.ts
+++ b/qubesmanager/i18n/qubesmanager_en.ts
@@ -636,7 +636,7 @@ p, li { white-space: pre-wrap; }
     <name>MainWindow</name>
     <message>
         <location filename="../ui_templatemanager.py" line="84"/>
-        <source>Template Manager</source>
+        <source>Template Switcher</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
@@ -1714,7 +1714,7 @@ The qube must be running to disable seamless mode; this setting is not persisten
     </message>
     <message>
         <location filename="../template_manager.py" line="66"/>
-        <source>Template Manager</source>
+        <source>Template Switcher</source>
         <translation type="unfinished"></translation>
     </message>
 </context>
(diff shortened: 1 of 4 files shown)


qubesmanager/i18n/qubesmanager_en.ts
```diff
@@ -679,7 +679,7 @@ p, li { white-space: pre-wrap; }
     </message>
     <message>
         <location filename="../utils.py" line="490"/>
-        <source>Whoops. A critical error has occured. This is most likely a bug in Qubes Manager.&lt;br&gt;&lt;br&gt;&lt;b&gt;&lt;i&gt;{0}&lt;/i&gt;&lt;/b&gt;&lt;br/&gt;at line &lt;b&gt;{1}&lt;/b&gt;&lt;br/&gt;of file {2}.&lt;br/&gt;&lt;br/&gt;</source>
+        <source>Whoops. A critical error has occured. This is most likely a bug in Qubes Switcher.&lt;br&gt;&lt;br&gt;&lt;b&gt;&lt;i&gt;{0}&lt;/i&gt;&lt;/b&gt;&lt;br/&gt;at line &lt;b&gt;{1}&lt;/b&gt;&lt;br/&gt;of file {2}.&lt;br/&gt;&lt;br/&gt;</source>
```
marmarek, line 682: This one is not that easy. First of all, "Qubes Switcher" is not a thing. But then, the `handle_exception` function is used by all the applications in this repo (template switcher, template manager, qubes manager, backup, backup restore). I'd either leave it as "Qubes Manager", or save the application name somewhere (or maybe it can be extracted from the QApplication singleton object?) and have it parametrized in this string.

marmarek (changes requested):

qubesmanager/i18n/qubesmanager_en.ts
alimirjamali: OK. Will look into it and fix it properly

alimirjamali: Passing additional parameter to `handle_exception` is possible within `run_synchronous` and `sys_asyncronous` (which is used here). But it would further complicate it. Since this is used most likely in debug scenarios, I think leaving it as "Qubes Manager" is the better solution as you suggested. Changes will be pushed in few seconds

marmarek (approved):

PR #371, opened 2024-07-16
Disable "Disposable template" checkbox for non-AppVMs

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/7217

### DIFF
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -897,6 +897,9 @@ class VMSettingsWindow(ui_settingsdlg.Ui_SettingsDialog, QtWidgets.QDialog):
         self.dvm_template_checkbox.setChecked(
             getattr(self.vm, 'template_for_dispvms', False))
 
+        if not hasattr(self.vm, 'template_for_dispvms'):
+            self.dvm_template_checkbox.setEnabled(False)
+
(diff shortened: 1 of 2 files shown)


qubesmanager/settings.py
```diff
@@ -897,6 +897,9 @@ def __init_advanced_tab__(self):
         self.dvm_template_checkbox.setChecked(
             getattr(self.vm, 'template_for_dispvms', False))
 
+        if getattr(self.vm, 'klass', False) != 'AppVM':
```
marmarek, line 900: Instead of hardcoding specific class, better check if it has the `template_for_dispvms` attribute. That will handle any future changes, but also a little know feature that StandaloneVM can be a template for disposable qube too.

marmarek (changes requested):

qubesmanager/settings.py
alimirjamali: > StandaloneVM can be a template for disposable qube too.

That is interesting to know. I wonder how Appmenu would behave if a Standalone is turned into a Disposable Template. And user could install software in the standalone if clicking on it in the Appmenu only opens a disposable based on it.

marmarek (approved):

PR #375, opened 2024-07-28
Unify qubes-os XDG config directory

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/8694
Removing space and upper-case letters from qubes-manager configuration directory.
Migrating old config to new unified config directory.

### DIFF
--- a/qubesmanager/utils.py
+++ b/qubesmanager/utils.py
@@ -32,6 +32,9 @@ import sys
 import qasync
 from qubesadmin import events
 from qubesadmin import exc
+import xdg.BaseDirectory
+import pathlib
+import shutil
 
 from PyQt5 import QtWidgets, QtCore, QtGui  # pylint: disable=import-error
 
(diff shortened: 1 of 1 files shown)


alimirjamali:
Making pylint happy(er)

marmarta:
I really like this idea. The "The Qubes Project" directory is I think a very old tech debt from something someone wrote 10 yrs ago :D

qubesmanager/qube_manager.py
```diff
@@ -715,9 +719,25 @@ class VmManagerWindow(ui_qubemanager.Ui_VmManagerWindow, QMainWindow):
...
+            '/The Qubes Project'
+        new_config_home = xdg.BaseDirectory.xdg_config_home + '/qubes-os'
+        config_file = 'qubes-qube-manager.conf'
+        old_config_file = old_config_home + '/' + config_file
+        new_config_file = new_config_home + '/' + config_file
+        if path.isfile(old_config_file) and not path.isfile(new_config_file):
+            pathlib.Path(new_config_home).mkdir(parents=True, exist_ok=True)
+            shutil.copy(old_config_file, new_config_file)
+
+        QCoreApplication.setOrganizationName('qubes-os')
```
marmarta, line 738: This actually duplicates code from utils.py (`run_asynchronous` and `run_synchronous` functions, where the general setup for QApplications is made. I think that file would be a better place for changes like this? And generally, maybe if we put sensible org name there, and the migration code also there, it could be only done in one place and not in multiple places (for other manager tools).

marmarta (changes requested):

qubesmanager/qube_manager.py
alimirjamali: It is done. The code is transferred to `utils.py` as advised and wrapped inside a function (so it would look nicer). The call to the said function is also done within the module.

alimirjamali:
Making pylint happy(er) since the indents were wrong

qubesmanager/qube_manager.py
alimirjamali: > I would propose to have this function be called in run synchronous/ run asynchronous helper functions so that when other tools from this package start using QSettings, it does not end in ressurection of The Qubes Project directory

It is done. Calls are from `run_synchronus` and `run_asynchronous` function.

> and conversely, to remove the old setting of app name from lines 576 and 536.

I wonder why I missed them. They are now deleted.

qubesmanager/utils.py
```diff
@@ -49,6 +52,25 @@
 # - just a list, no properties or defaults, just a nice list with a "current"
 #       value - initialize_widget
 
+def set_organization():
+    """set the organization and config home directory for Qt based Apps."""
+    # migrate old config file to the unified config directory
+    # TODO: remove the below block when Qubes R4.3 becomes End-of-Life
+    old_config_home = xdg.BaseDirectory.xdg_config_home + \
+        '/The Qubes Project'
+    new_config_home = xdg.BaseDirectory.xdg_config_home + '/qubes-os'
```
ben-grande, line 61: I believe it is better to derive the new configuration directory from the `OrganizationName` to avoid code duplication.

qubesmanager/utils.py
```diff
@@ -49,6 +52,25 @@
...
+    config_file = 'qubes-qube-manager.conf'
```
ben-grande, line 62: I believe it is better to derive the new configuration file from the `ApplicationName` to avoid code duplication.

ben-grande (changes requested):
Minor changes requested. Looks good. Untested.

qubesmanager/utils.py
alimirjamali: This is not possible. Since the config file copy operation (if needed) happens before the QtCore initialization. This is done intentionally to avoid any conflict with QtCore mechanism. Just to be on the safe side.

qubesmanager/utils.py
alimirjamali: Same as above comment

marmarta (approved):

PR #385, opened 2024-09-03
Remove stale, incomplete QubeManager release notes

alimirjamali:
resolves: https://github.com/QubesOS/qubes-issues/issues/8803


marmarta (approved):
I think this is a good idea. And yeah, we can add release notes to global config together with various other status info.

PR #386, opened 2024-09-11
Fixes to Template Switcher

marmarta:
- improved behavior on double-click
- improved behavior on OK (progress dialog)
- no auto-close on OK

fixes QubesOS/qubes-issues#7639

### DIFF
--- a/qubesmanager/common_threads.py
+++ b/qubesmanager/common_threads.py
@@ -72,3 +72,23 @@ class CloneVMThread(QubesThread):
             self.msg_is_success = True
         except exc.QubesException as ex:
             self.msg = (self.tr("Error while cloning qube!"), str(ex))
+
+
+class ChangeTemplatesThread(QtCore.QThread):
+    def __init__(self, progress_dialog, items_to_change, qubes_app):
+        super().__init__()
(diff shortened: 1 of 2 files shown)


qubesmanager/common_threads.py
```diff
@@ -72,3 +72,24 @@ def run(self):
...
+        self.dialog = progress_dialog
+        self.items = items_to_change
+        self.qubes_app = qubes_app
+        self.errors = {}
+
+    def run(self):
+        i = 0
+        for vm, row in self.items:
+            i += 1
+            if row.new_item and row.new_item.changed:
```
marmarek, line 89: `items_to_change` is already filtered on this condition

marmarek:
> no auto-close on OK

Shouldn't the button be named "apply" then? Right now it's confusing that the app remains open when everything got applied successfully (yes, I know, it's the exact opposite to what issue description said :P ).

marmarek (changes requested):

marmarek:
Oh, and when changing button to "apply", maybe change also "cancel" to "close"? Because currently it may suggest that "cancel" would rollback the change...

marmarek (approved):

PR #392, opened 2024-11-05
Template Manager GUI: better info for upgradeable templates

marmarta:
If upgrade comes from different repository than the installed template, show the repository in description.

### DIFF
--- a/qubesmanager/qvm_template_gui.py
+++ b/qubesmanager/qvm_template_gui.py
@@ -164,6 +164,8 @@ class Template(TreeItem):
         self.system_eol_date = None
         self.newest_available_version = entry.get('upgraded_version',
                                                   self.version_release)
+        self.newest_version_repo = entry.get('upgraded_repo',
+                                             self.repository_name)
 
     @property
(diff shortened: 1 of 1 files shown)


qubesmanager/qvm_template_gui.py
```diff
@@ -184,7 +186,10 @@ def description(self) -> str:
             text += tr("<b>Template upgrade available: </b>")
             if self.template_status == 'upgradable':
                 text += tr("yes, newest available version: " +
-                           self.newest_available_version + "<br>")
+                           self.newest_available_version)
```
marmarek, line 189: It was an issue before, but variable part should not be under the first arg to `tr()`

marmarek (changes requested):

qubesmanager/qvm_template_gui.py
marmarta: fixed

marmarek (approved):

marmarek:
PipelineRetry

PR #393, opened 2024-11-09
Retire old `qubes-vm-create` dialog

alimirjamali:
resolves: https://github.com/QubesOS/qubes-issues/issues/6561


alimirjamali:
Development notes: Adding `python3-xdg` to requirements to let the CI/CD tests run. Dependency was introduced during  [#375](https://github.com/QubesOS/qubes-manager/pull/375)

marmarek:
As noted in the other issue, IMO better to simply remove the old create qube dialog (and replace its use with the new one).

marmarek:
I'll pick just the CI fix from here

alimirjamali:
> As noted in the other issue, IMO better to simply remove the old create qube dialog (and replace its use with the new one).

OK. But I think it is better to do it for 4.3 branch and not 4.2. I will make a separate PR for the above properly ditching the old one. So it could be easily merged to 4.3 branch.

qubesmanager/qube_manager.py
```diff
@@ -1398,10 +1400,16 @@ def update_network_menu(self):
...
     @pyqtSlot(name='on_action_createvm_triggered')
     def action_createvm_triggered(self):
-        with common_threads.busy_cursor():
-            create_window = create_new_vm.NewVmDlg(
+        if self.new_qube.state() == QProcess.ProcessState.Running:
+            return
+        if path.isfile("/usr/bin/qubes-new-qube"):
+            self.new_qube.start("/usr/bin/qubes-new-qube")
+        else:
+            # for stubborn users who might uninstall `desktop-linux-manager`
```
marmarta, line 1408: Given how much functionality is there, I think that if a user manually uninstalls desktop-linux-manager we can safely leave them on their own.... Same as with qubes-global-config, if a user uninstalls a system package manually, some things will indeed fail. I think we can just remove the old create_new_vm dialog in its entirely and call here the `qubes-new-qube` dialog.

marmarta (changes requested):

qubesmanager/qube_manager.py
alimirjamali: OK. So lets go for the kill and delete `create_new_vm.py` and the related ui file. I will readjust the PR.

alimirjamali: OK. This is done properly. The main executable, desktop file, unittest, ui file are deleted. Packaging files, project files, translations, Makefile and setup file are updated.

marmarta (approved):

marmarek:
Now that #400 is merged, this needs rebase...

marmarek (approved):

PR #394, opened 2024-11-13
Refresh disk usage stats every 60s if focused

alimirjamali:
resolves: https://github.com/QubesOS/qubes-issues/issues/7619

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -876,6 +876,8 @@ class VmManagerWindow(ui_qubemanager.Ui_VmManagerWindow, QMainWindow):
         dispatcher.add_handler('domain-feature-delete:updates-available',
                                self.on_domain_updates_available)
 
+        self.installEventFilter(self)
+
         # It needs to store threads until they finish
         self.threads_list = []
         self.progress = None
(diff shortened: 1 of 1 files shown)


marmarek:
It looks like it doesn't do what it says it does. It updates every 60s when active.
The cache remains valid for 30s after returning the data. The timer triggers after 30s after scheduling previous call (which only latter fetched the usage stats). So, I think every second call still hits the cache.

alimirjamali:
> It looks like it doesn't do what it says it does. It updates every 60s when active. The cache remains valid for 30s after returning the data. The timer triggers after 30s after scheduling previous call (which only latter fetched the usage stats). So, I think every second call still hits the cache.

I think it is better to increase the timer value to 60s. To assure that it is not hitting any cache. And it does exactly what the commit message says. Overall this is a very minor issue.

The amendment will follow in few  seconds.

marmarek (approved):

marmarek:
I tried backporting it to release4.2 branch. There is a conflict trivial to resolve, but then it doesn't work. Can you help? It's likely qt5 vs qt6 (release4.2 remains at qt5).

I also noticed manager showing absurd sizes for running VMs vs stopped ones. For example "personal" qube takes below 100MiB when stopped, but just starting it makes it over 6GiB, then stopping goes back to below 100MiB. I don't see how this commit could result in such a behavior, weird.

alimirjamali:
> I tried backporting it to release4.2 branch. There is a conflict trivial to resolve, but then it doesn't work. Can you help? It's likely qt5 vs qt6 (release4.2 remains at qt5).

I will look in to this.

> I also noticed manager showing absurd sizes for running VMs vs stopped ones. For example "personal" qube takes below 100MiB when stopped, but just starting it makes it over 6GiB, then stopping goes back to below 100MiB. I don't see how this commit could result in such a behavior, weird.

The old behaviour should be similar and unrelated to this commit. It should just takes longer (i.e. up to 5 minutes) to show. And the reason is [here](https://github.com/QubesOS/qubes-manager/blob/main/qubesmanager/qube_manager.py#L357-L363). Where the `get_disk_utilization` is used. The problem is not within `qubes-manager`. It is within `qubesadmin` where it calculates the sum of all volumes (including `root-dirty.img`) even for AppVMs [here](https://github.com/QubesOS/qubes-core-admin-client/blob/main/qubesadmin/vm/__init__.py#L264-L266).

marmarek:
I forgot this system is using btrfs, looks to be an issue with file-reflink driver (or at least related to it).

alimirjamali:
> I forgot this system is using btrfs, looks to be an issue with file-reflink driver (or at least related to it).

My system is BTRFS based and I experience the exact symptom

PR #399, opened 2024-11-30
Add filter for qubes stored on inaccessible storage pools

Ofenhed:
This is useful when you're working for multiple customers and compartmentalize them to separate (removable?) disks. If you decide that this can be merged, I'd also like to add the same filter to `qubes-app-menu` as well.

This is slightly related to [QubesOs/qubes-issues#1293](https://github.com/QubesOS/qubes-issues/issues/1293), as this is part of a work flow where VMs can be stored on another partition and encrypted with a separate key, as described [here](https://www.qubes-os.org/doc/secondary-storage/). Of course, this PR doesn't add any security, as all metadata is still stored on the same place. It's purely a visual improvement if users have themselves implemented the separation.

For context, the way I'm using this is with a [salt script](https://github.com/Ofenhed/qubes-salts/blob/main/external-disk.sls) which adds some safe guards to avoid corrupted data.


Ofenhed:
I think the reason this works and updates is because Qubes periodically queries for disk space usage. Should this use the event system, instead of polling for volumes?

There is a quirk in this implementation, if you do weird setups where you keep AppVMs separate from their TemplateVMs, but the TemplateVM isn't on main storage, as it currently does not verify that the template volume of an AppVM is accessible. This would result in false positives where the AppVM is visible, even though it cannot be started.

marmarta:
I really like the idea! Don't love the polling, for two reasons:
- it can be pretty resource-intense
- the way it's done, it breaks a bit the separation we have between model and display data. If you look at the manager code, we try to keep the VM-relevant API calls in the `VMInfo` class - it would be better to put the logic there, and update it using the events system; if you need advice on which events and where to find them, tell me here (if you know it well, I don't want to be explaining things you know to you :) )

Ofenhed:
Thank you for the positive feedback. I agree about the polling, I just did what felt manageable to get some feedback on the idea.

I'd gladly take some advice. My understanding right now is that there would have to be polling, but that I could use the polling already happening in `qubes-core-admin`. As far as I can tell there is currently no events implemented, so I'd have to add events ("pool-added", "pool-removed"?) in `qubes-core-admin/qubes/storage/`. Is this correct? I haven't found a way to do it yet, outside of implementing change detection separately for each Pool backend. It seems that I might be able to use `Volume` for added volumes, but I haven't tried yet, and I haven't found a way to detect removed devices. It seems easy if they are willfully removed via API calls, but I haven't found a general way of detecting pools lost because their backend (such as an LVM group dissapearing) was removed.

marmarta:
Sorry for the late reply, Christmas season was a bit overwhelming.

Anyway.
- if you want to add new events, the events need to go in both `core-admin` and `core-admin-client` - `core-admin` is the backend who throws actual events, `core-admin-client` is the thing frontend apps use and this just passes the event forward - you can check existing events like `property-set` to see how it's done; those repos like a lot of tests, otherwise @marmarek is unhappy
- so, there is a new device API happening (PR here: https://github.com/QubesOS/qubes-core-admin/pull/579) - maybe this can be tied to existing there events like `device-list-change:block` (I think block...) instead of adding the logic to pools? OTOH I imagine there could be changes in pools that are not related to adding/removing devices (although this does boggle and terrify the mind a bit, at least my mind, when I think about editing existing storage pools while using them)

marmarek:
> I'd gladly take some advice. My understanding right now is that there would have to be polling, but that I could use the polling already happening in `qubes-core-admin`. As far as I can tell there is currently no events implemented, so I'd have to add events ("pool-added", "pool-removed"?) in `qubes-core-admin/qubes/storage/`. Is this correct? I haven't found a way to do it yet, outside of implementing change detection separately for each Pool backend. It seems that I might be able to use `Volume` for added volumes, but I haven't tried yet, and I haven't found a way to detect removed devices. It seems easy if they are willfully removed via API calls, but I haven't found a general way of detecting pools lost because their backend (such as an LVM group dissapearing) was removed.

There may be already events when a new pool is created, but not really when a pool becomes available/unavailable. The latter in practice requires some polling anyway, as it may change outside of knowledge of qvm-* tools, for example when you mount something in dom0, or open related LUKS volume (theoretically some of the cases could be covered by udev events, but it's hard to map events to pools, and still probably some will be missing).
And with polling needed anyway, I'd rather avoid it if nobody uses it at the time - with events, we don't really know (at the core-admin side) what events are listened for, so there is no way to enable polling only when some app is waiting for pool-specific events.

So, as long as polling is needed, unfortunately it seems it needs to be on the qubes-manager side...

And BTW, events for disk usage would be even more tricky, as it needs to be some polling anyway, or at least rate-limiting (the non-polling version would be "send event on any disk write" which is totally impractical). And on top of that, has the same issues as pools availability (no system events on changes, no info whether anybody listens, etc).

marmarek:
PipelineRetry

Ofenhed:
I've noticed an issue with this, where I have a standalone VM which only shows up when I disable the filtering.

marmarek:
I tried to test this and got a weird behavior:
1. First of all, unchecking the checkbox did not hide a qube from unavailable storage pool.
2. But it did hide almost all other qubes. Only running qubes remained, plus dom0, plus the one on unavailable storage pool...

Is there maybe some condition reversed?

Ofenhed:
You were right, there was a condition reversed. There was also a pretty bad performance issue I wasn't noticing before. I refactored a bit to fix it in the same way as disk usage is updated.
 
I reworked it a bit, but it's still a draft with a lot of failed tests, but I'm afraid I don't have time to finish it right now.

qubesmanager/qube_manager.py
```diff
@@ -341,6 +368,9 @@ def update(self, update_size_on_disk=False, event=None):
                 self.disk_float = None
                 self.disk = None
 
+        if self.vm.klass != 'AdminVM' and update_availability:
```
Ofenhed, line 371: Should probably set `available` to true without checking volumes, if the VM is running.

qubesmanager/qube_manager.py
```diff
@@ -1139,6 +1192,14 @@ def update_running_size(self, *_args):
                 self.qubes_cache.get_vm(qid=vm.qid).update(
                     update_size_on_disk=True, event='disk_size')
 
+    def update_halted_availability(self, *_args):
+        if self.show_unavailable_pool_action.isChecked():
```
Ofenhed, line 1196: Note to self: Invert 🤦

qubesmanager/qube_manager.py
```diff
@@ -1139,6 +1192,14 @@ def update_running_size(self, *_args):
...
+            for vm in self.qubes_app.domains:
+                if not vm.is_running():
+                    self.qubes_cache.get_vm(qid=vm.qid).update(
+                        update_availability=True, event='volume_availability')
+            self.invalidate()
```
Ofenhed, line 1201: This feels wrong to put here, but the UI did not seem to update without it.

qubesmanager/qube_manager.py
```diff
@@ -227,9 +227,31 @@ def __init__(self, vm):
...
+            if volume.ephemeral:
+                continue
+            if volume.pool not in self.vm.app.pools:
+                return False
+            if volume.source is not None:
+                if volume.source not in (
+                    pool_volume.vid
+                    for pool in self.vm.app.pools.values()
+                    for pool_volume in pool.volumes
+                ):
```
Ofenhed, line 246: This check seems kind of expensive, but I don't know if there is a simpler method that's also backend agnostic.

qubesmanager/qube_manager.py
```diff
@@ -1139,6 +1192,14 @@ def update_running_size(self, *_args):
                 self.qubes_cache.get_vm(qid=vm.qid).update(
                     update_size_on_disk=True, event='disk_size')
 
+    def update_halted_availability(self, *_args):
+        if not self.show_unavailable_pool_action.isChecked():
```
Ofenhed, line 1196: Checking the box should probably trigger this function.

Ofenhed:
I noticed that I reinvented the wheel. I changed to use the `volume.validate()` function instead, and performance increased dramatically. It should also be more consistent with actual availability as it removes my assumptions.

Ofenhed:
I think I'll drop this pull request in favor of something more generic and useful for a broader part of the user base. Implementing a tag filter in qubes-manager pretty much fills the same function. It's not as easy to use, as it's no longer automatic, but it's a lot more versatile. Related issue for the same functionality in the app laucher: [QubesIssues#7768](https://github.com/QubesOS/qubes-issues/issues/7768).

PR #406, opened 2025-02-16
Add free-form text to qube for notes, comments, ...

alimirjamali:
qubesadmin part of adding free-form text to each qube for comments, notes, descriptions, remarks, reminders, etc.

fixes: https://github.com/QubesOS/qubes-issues/issues/899


alimirjamali:
@marmarta This is ready for review.

It requires https://github.com/QubesOS/qubes-core-admin-client/pull/334

marmarta:
I like how it looks, one thing I'd love to see: adding a label with explanation to the tab. Something like this above the text field:
"Notes and comments
For user reference only"

Or something similar, if you have better ideas on how to formulate it. I'd like to make it clear for users this is not in any way used by Qubes backend itself.

alimirjamali:
I believe we have 4 options.

1. Above comment text field.
2. Below text field.
3. As a toolTip
4. As a place holder text (explanation could be much longer in this case e.g size limit)

Or a mixture of the above. I personally prefer place holder or below text field. I appreciate if you could advise. Screenshots for the choices is attached below

notes_tips

marmarta:
I really like the placeholder idea - I'm not sure if "start typing to erase this placeholder text" is necessary, though? How do you think.

alimirjamali:
> I really like the placeholder idea - I'm not sure if "start typing to erase this placeholder text" is necessary, though? How do you think.

I agree. The suggested change is applied.

marmarek:
PipelineRetryFailed

marmarek (approved):

PR #407, opened 2025-02-17
Add boot mode support

ArrayBolt3:
Adds the user interface elements described at https://github.com/QubesOS/qubes-issues/issues/9750#issuecomment-2655149125 to the settings UI.

Not fully tested, thus not ready for merge. It seems to work for the most part, but my PR for the backend has a problem that keeps me from changing AppVM's templates for some reason.

### DIFF
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -933,19 +933,6 @@ class VMSettingsWindow(ui_settingsdlg.Ui_SettingsDialog, QtWidgets.QDialog):
                     default_text_provider=get_default_bootmode_name
                 )
                 if hasattr(self.vm, "appvm_default_bootmode"):
-                    # We need to recreate the bootmode_widget_data list,
-                    # because utils.initialize_widget_for_property adds a
-                    # default item to the list as a side-effect, and we'd end
-                    # up with duplicate "default" entries if we didn't
-                    # reinitialize it. Modifying
-                    # initialize_widget_for_property to operate on a list deep
-                    # copy breaks the virtualization mode combo box, making
-                    # the default mode appear as
-                    # `<object object at 0xaddress>`.
-                    bootmode_widget_data = list(zip(
-                        self.bootmode_names, self.bootmode_ids
-                    ))
-                    bootmode_widget_data.sort()
                     utils.initialize_widget_for_property(
                         widget=self.appvm_default_bootmode,
                         choices=bootmode_widget_data,
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -63,6 +63,16 @@ INTERNAL_SERVICE_FEATURES = [IDLE_SERVICE]
 INTERNAL_SUPPORTED_FEATURES = [IDLE_SUPPORTED_SERVICE]
 
 
+def get_default_bootmode_name(vm, bootmode):
+    if bootmode == "default":
+        return vm.features.check_with_template(
+            "boot-mode.name.default", ""
+        )
+    return vm.features.check_with_template(
+        f"boot-mode.name.{bootmode}", bootmode
+    )
+
+
 # pylint: disable=too-few-public-methods
 class RenameVMThread(common_threads.QubesThread):
     def __init__(self, vm, new_vm_name, dependencies):
@@ -392,6 +402,25 @@ class VMSettingsWindow(ui_settingsdlg.Ui_SettingsDialog, QtWidgets.QDialog):
 
         self.netvm_no_firewall_label.setVisible(no_firewall_state)
         self.sysnet_warning_label.setVisible(netvm is None and provides_network)
+
+    def collect_bootmode_data(self):
+        bootmode_ids = [
+            x.split('.')[2] for x in self.vm.features \
+                if x.startswith("boot-mode.kernelopts.")
+        ]
+        subject = self.vm
+        while hasattr(subject, "template"):
+            bootmode_ids.extend([
+                x.split('.')[2] for x in subject.template.features \
+                    if x.startswith("boot-mode.kernelopts.")
+            ])
+            subject = subject.template
+        bootmode_names = [
+            self.vm.features.check_with_template(
+                f"boot-mode.name.{x}", x
+            ) for x in bootmode_ids
+        ]
+        return bootmode_names, bootmode_ids
 
     def current_tab_changed(self, idx):
         if idx == self.tabs_indices["firewall"]:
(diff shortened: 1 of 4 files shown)


ArrayBolt3:
Figured out the bug. From a UX standpoint this seems done at least when doing a quick smoke test. Will do more thorough testing on this and https://github.com/QubesOS/qubes-core-admin/pull/653 and hopefully mark as ready for review tomorrow.

ArrayBolt3:
This is now ready for review. This should be merged with the corresponding backend PR, https://github.com/QubesOS/qubes-core-admin/pull/653.

ArrayBolt3:
Sigh, my Git acrobatics seems to have confused GitHub...

qubesmanager/settings.py
```diff
@@ -882,6 +892,69 @@ def __init_advanced_tab__(self):
...
+                    bootmode_widget_data.sort()
+                    utils.initialize_widget_for_property(
+                        widget=self.appvm_default_bootmode,
+                        choices=bootmode_widget_data,
+                        property_name="appvm_default_bootmode",
+                        holder=self.vm,
+                        allow_default=True,
+                        default_text_provider=get_default_bootmode_name
+                    )
+                self.bootmode_kernel_opts.setText(
```
marmarek, line 952: From the other PR - can you add debug print to see `self.vm.bootmode` value here?

ArrayBolt3: I did that and saw that it was printed properly. I was able to fix this issue though using your event handling trick, I just ended up having to use a different event than the one you initially suggested. (I'm not entirely sure `feature-set` exists in the context of qubesvm.py, it seems only `domain-feature-set` does.)

ArrayBolt3:
Just fixed a label formatting issue.

marmarek:
As a side effect you made the window quite tall. The basic one has a lot of padding at the bottom now, and similarly even on the advanced tab too. And the buttons don't fit on (small) screen: https://openqa.qubes-os.org/tests/131086#step/guivm_manager/27

ui/settingsdlg.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>836</width>
-    <height>665</height>
+    <width>873</width>
+    <height>840</height>
```
marmarek, line 10: I think it's simply this change. Revert to the old value (or slightly increase if needed in advanced tab).

ArrayBolt3: Ah, that would make sense. I bet what happened is me opening the file in Qt Designer on my Kubuntu machine caused everything to get taller (and probably wider too) because of the size of Breeze widgets vs. the size of widgets when using Qt on XFCE. If I look at it in Designer with Breeze, it looks perfectly fine, but all the buttons are a different size here too.

I tried forcing the use of Fusion and it seems to have made things a bit smaller... but not by much. I guess I need to find out what theme is being used in Qubes OS itself and use that (or just develop on my Qubes OS machine rather than my main development rig).

marmarek: Maybe simply try to manually edit those two to the old values?

ArrayBolt3:
@marmarek I tried to fix the pylint issues, which seems to have made it happy, but the regression tests now fail:

```
======================================================================
ERROR: qubes.tests.vm.qubesvm/TC_90_QubesVM/test_811_default_bootmode
----------------------------------------------------------------------
  File "/usr/lib64/python3.13/contextlib.py", line 85, in inner
    return func(*args, **kwds)
  File "/home/user/qubes-core-admin/qubes/tests/vm/qubesvm.py", line 3121, in test_811_default_bootmode
    vm.features["boot-mode.active"] = "testmode1"
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/home/user/qubes-core-admin/qubes/features.py", line 93, in __setitem__
    self.subject.fire_event(
    ~~~~~~~~~~~~~~~~~~~~~~~^
        "domain-feature-set:" + key, feature=key, value=value
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/user/qubes-core-admin/qubes/events.py", line 200, in fire_event
    sync_effects, async_effects = self._fire_event(
                                  ~~~~~~~~~~~~~~~~^
        event, kwargs, pre_event=pre_event
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/user/qubes-core-admin/qubes/events.py", line 169, in _fire_event
    effect = func(self, event, **kwargs)
TypeError: QubesVM.on_feature_bootmode_active_set() missing 1 required positional argument: 'event'
```

I can't figure out why this happens though - `qubes/vm/qubesvm.py -> on_feature_bootmode_active_set` uses this "signature" (not sure what these are called in Python):

```
    @qubes.events.handler("domain-feature-set:boot-mode.active")
    def on_feature_bootmode_active_set(
        self, vm, event, feature, value, oldvalue=None
    ):
```

This is nearly identical to a signature in `qubes/ext/services.py -> on_domain_feature_set`:

```
    @qubes.ext.handler("domain-feature-set:*")
    def on_domain_feature_set(self, vm, event, feature, value, oldvalue=None):
```

How come the second one works but the first one doesn't? (The first one used to work without problems before adding `self` as the first argument, as required by `pylint`.)

ui/settingsdlg.ui
ArrayBolt3: yep, I will, just trying to figure out how to avoid this happening in the future.

ArrayBolt3: Tested, that fixed it.

marmarek:
If you change the boot mode, the "Boot mode kernel opts" entry doesn't update immediately, only after clicking "Apply". This is IMO confusing.

As for the failing tests, mockups needs to be taught the new properties. That can be done in https://github.com/QubesOS/qubes-core-admin-client/blob/main/qubesadmin/tests/mock_app.py#L109. And while at it, document also new properties and features in relevant man pages (which is also in core-admin-client repo).
And also, this new section want to have unit test.

ArrayBolt3:
> If you change the boot mode, the "Boot mode kernel opts" entry doesn't update immediately, only after clicking "Apply". This is IMO confusing.

In retrospect I agree, I did notice that somewhat but didn't think much of it since I was implementing the label about the same way the other kernelopts label was implemented. But this functions in a different way since it's easily user-changeable, so that's worth resolving.

Does the code look OK aside from needing tests still? I'll work on those tonight.

marmarek:
Overall it looks fine for me, maybe extract enumerating boot modes (constructing `bootmode_widget_data`) to a helper function to not make the init tab function even longer.

ArrayBolt3:
Found some bugs while writing the regression tests, took me a while to figure out how to use the rtest engine just right but I finally got it, I think. I still need to break out the code you mentioned into a new function and re-review what I just wrote to make sure there aren't any glaring errors.

ArrayBolt3:
Self-review done, found and fixed some minor issues in qubes-manager and one issue that snuck by me in qubes-core-admin.

marmarek:
But you can have a VM with non-default boot mode, no? Just set the `bootmode` property in the mockup (probably instead of setting `boot-mode.active` feature, which isn't relevant to the frontend).

ArrayBolt3:
The value of the `bootmode` property will be ignored when using `property_get_default` though, because we're getting the property's default value so that we have something to fill in for the `default (name)` selection in the combo box. In actual use, that default value would come from the `boot-mode.active` feature if possible, but that doesn't work in the tests for obvious reasons.

marmarek:
I mean the other not covered line. The one about `property_get_default` is less important.

ArrayBolt3:
ah, I see what you mean now (I had it in my head that both were in a similar situation but I see now that's not the case). I'll add another mockup and another testcase and see if I can get that one too.

ArrayBolt3:
@marmarek I believe the addition to the tests you wanted is now added. `coverage html` shows the `if self.vm.bootmode != "default":` bit on settings.py line 957 is now covered

qubesmanager/settings.py
```diff
@@ -882,6 +911,59 @@ def __init_advanced_tab__(self):
...
+                    allow_default=True,
+                    default_text_provider=get_default_bootmode_name
+                )
+                if hasattr(self.vm, "appvm_default_bootmode"):
+                    # We need to recreate the bootmode_widget_data list,
+                    # because utils.initialize_widget_for_property adds a
+                    # default item to the list as a side-effect, and we'd end
+                    # up with duplicate "default" entries if we didn't
+                    # reinitialize it. Modifying
+                    # initialize_widget_for_property to operate on a list deep
```
marmarta, line 941: I'm not sure I get this - maybe this should not be recreating the list, but an `hasattr(self.vm, "appvm_default_bootmode")` ... `else` ...?

marmarta (changes requested):

qubesmanager/settings.py
ArrayBolt3: The problem is that when you use `utils.initialize_widget_for_property(... choices=bootmode_widget_data allow_default=True ...)`, the function will actually modify the variable passed in through the `choices` argument, adding a `default` entry to the list. This modifies the real variable, it doesn't work on a copy of the variable. So if you pass the same list to `initialize_widget_for_property` twice, with `allow_default=True` set for both calls, the second widget you initialize will have *two* `default` entries in it. We need to initialize both the `bootmode` and the `appvm_default_bootmode` widgets with the same list, so we have to create the list twice (or remove the `default` option from the list before initializing the second widget, the former seems more reliable to me).

The ideal solution in my mind would be to make the `initialize_widget_for_property` call work on a deep copy of the `choices` list, rather than modifying it in-place. I originally implemented that, and it works for the most part, but it broke the `virt_mode` combo box in the way described in the comment. That indicated to me that some things depended on this seemingly broken behavior from `initialize_widget_for_property`, so I went back on that since I didn't want to break things or change more than I had to in this PR.

I'm not sure what would go in the `else:` block with your suggestion. Could you elaborate?

marmarta: ok, I see - I think I misunderstood it at first read. Still not that much of a fan, it does look clunky, so if you think fixing the `initialize_widget_for_property`/`virt_mode` behavior is doable (in a separate commit, so marmarek is happy :D), I think it would be better?

ArrayBolt3: I'll try to do that.

ArrayBolt3: Done, turns out a shallow copy was enough.

marmarek (approved):
LGTM; but I'd like also @marmarta confirmation the interface is ok

marmarta (approved):

PR #408, opened 2025-02-17
Add the feature to prohibit starting a qube

alimirjamali:
Qube Manager part of preventing qubes with `prohibit-start` from being started. Showing special status icon and disabling start and related buttons.

Supplements: https://github.com/QubesOS/qubes-issues/issues/9622

### DIFF
--- a/icons/ban.svg
+++ b/icons/ban.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-ban"><circle cx="12" cy="12" r="10" stroke="#404040"/><path d="m4.9 4.9 14.2 14.2" stroke="#404040"/></svg>
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -100,7 +100,8 @@ class StateIconDelegate(QStyledItemDelegate):
                 "Transient" : QIcon(":/transient"),
                 "Halting" : QIcon(":/transient"),
                 "Dying" : QIcon(":/transient"),
-                "Halted" : QIcon(":/blank")
+                "Halted" : QIcon(":/blank"),
+                "Blocked" : QIcon(":/ban"),
                 }
         self.outdatedIcons = {
                 "update" : QIcon(":/updateable"),
@@ -193,8 +194,17 @@ class StateIconDelegate(QStyledItemDelegate):
             # sometimes it's not enough to use an empty string
             if index != self.lastIndex:
                 QToolTip.showText(QPoint(), ' ')
-            QToolTip.showText(event.globalPos(),
-                index.data()['power'], view)
+            if index.data()['power'] == 'Blocked':
+                QToolTip.showText(event.globalPos(),
+                    self.tr(
+                        "The qube is prohibited from starting\n"
+                        "See `qvm-features` manual for more information"
+                    ),
+                    view
+                )
+            else:
+                QToolTip.showText(event.globalPos(),
+                    index.data()['power'], view)
         else:
             margin = iconRect.left() - option.rect.left()
             left = delta = margin + iconRect.width()
@@ -234,6 +244,14 @@ class VmInfo():
     def update_power_state(self):
         try:
             self.state['power'] = self.vm.get_power_state()
+            if self.state['power'] == "Halted" and \
+                    self.vm.klass != "AdminVM" and \
+                    manager_utils.get_feature(
+                        self.vm,
+                        'prohibit-start',
+                        False
+                    ):
+                self.state['power'] = 'Blocked'
         except exc.QubesDaemonAccessError:
             self.state['power'] = ""
 
@@ -264,6 +282,8 @@ class VmInfo():
                     eol = datetime.strptime(eol_string, '%Y-%m-%d')
                     if datetime.now() > eol:
                         self.state['outdated'] = 'eol'
+                else:
+                    self.state['outdated'] = None
         except exc.QubesDaemonAccessError:
             pass
 
@@ -856,6 +876,10 @@ class VmManagerWindow(ui_qubemanager.Ui_VmManagerWindow, QMainWindow):
         dispatcher.add_handler('domain-shutdown', self.on_domain_status_changed)
         dispatcher.add_handler('domain-paused', self.on_domain_status_changed)
         dispatcher.add_handler('domain-unpaused', self.on_domain_status_changed)
+        dispatcher.add_handler('domain-feature-set:prohibit-start',
+                               self.on_domain_status_changed)
+        dispatcher.add_handler('domain-feature-delete:prohibit-start',
+                               self.on_domain_status_changed)
 
         dispatcher.add_handler('domain-add', self.on_domain_added)
         dispatcher.add_handler('domain-delete', self.on_domain_removed)
@@ -1130,6 +1154,10 @@ class VmManagerWindow(ui_qubemanager.Ui_VmManagerWindow, QMainWindow):
                 elif manager_utils.get_feature(
                         info.vm, 'updates-available', False):
                     info.state['outdated'] = 'update'
+                else:
+                    info.state['outdated'] = None
+            else:
+                info.state['outdated'] = None
         except exc.QubesDaemonAccessError:
             return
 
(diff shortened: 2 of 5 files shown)


alimirjamali:
Improving unittests and adding individual unittest. Should make the Codecov green

marmarek:
> [2025-02-21T12:06:32.376Z] ['info'] Pinging Codecov: https://codecov.io/upload/v4?package=uploader-0.8.0&token=*******&branch=pr-408&build=9207956754&build_url=&commit=00dbb84c06cec4bf0a9484102beb1a1dac425a55&job=&pr=408&service=gitlab&slug=QubesOS%2Fqubes-manager&name=&tag=&flags=&parent=
[2025-02-21T12:06:32.552Z] ['error'] There was an error running the uploader: Error uploading to https://codecov.io/: Error: There was an error fetching the storage URL during POST: 503 - upstream connect error or disconnect/reset before headers. reset reason: connection termination

I'll click restart...

qubesmanager/tests/test_qube_manager.py
```diff
@@ -1604,3 +1604,50 @@ def test_704_check_later(mock_timer, mock_question):
...
+    qubes_manager.dispatcher.add_expected_event(
+        MockEvent('test-standalone',
+                  'domain-feature-set',
+                  [('name', 'prohibit-start'),
+                   ('newvalue', 'Do not update')]))
+
+    with contextlib.suppress(asyncio.TimeoutError):
+        await asyncio.wait_for(qubes_manager.dispatcher.listen_for_events(), 1)
+
+    qubes_manager.qubes_app.domains['test-old'].features[ \
```
marmarta, line 1645: So, this should not happen here. The setting of features on the test objects does not actually change their state in any meaningful way - it only fires the appropriate event for Feature.Set, but does not change what happens for features_get. You need to set the state of everything - including setting those properties - before the update_vm_calls and firing up asyncio, and treat the firing of asyncio.listed_for_events as the moment those events actually happen. Setting the feature here will not change the state of the app.

alimirjamali: Thank you very much for the hint. The tests are all green now. I am still not fully certain if there is any extraneous parts here that has to be deleted. Otherwise this PR is ready.

icons/ban.svg
marmarta, line 1: So, the icon is good, but the problem with making it black is that it's going to completely vanish in dark mode. Dark mode is not fully officially supported, sure, but people do use it and I don't want to break it for them. What about #404040 ? It will be still dark and look appropriate but also be not invisible on black background.

qubesmanager/qube_manager.py
```diff
@@ -103,13 +103,16 @@ def __init__(self):
...
+                "blocked" : QIcon(":/ban"),
                 "update" : QIcon(":/updateable"),
                 "outdated" : QIcon(":/outdated"),
                 "to-be-outdated" : QIcon(":/outdated"),
                 "eol": QIcon(':/warning'),
                 "skipped": QIcon(':/skipped')
                 }
         self.outdatedTooltips = {
+                "blocked" : self.tr(
+                    "The qube is prohibited from being started"),
```
marmarta, line 115:
```suggestion
                    "The qube is prohibited from starting"),
```

(to avoid passive voice when it's not inconvenient)

qubesmanager/qube_manager.py
```diff
@@ -253,6 +262,9 @@ def update_power_state(self):
 
             if self.vm.klass in {'TemplateVM', 'StandaloneVM'}:
                 if manager_utils.get_feature(
+                        self.vm, 'prohibit-start', False):
+                    self.state['outdated'] = 'ban'
```
marmarta, line 266: should this be 'ban' and not 'blocked'? Also, this line is not covered by tests

qubesmanager/qube_manager.py
```diff
@@ -1352,6 +1375,17 @@ def table_selection_changed(self):
             if not vm.updateable and vm.klass != 'AdminVM':
                 self.action_updatevm.setEnabled(False)
 
+            if vm.vm.features.get('prohibit-start', False):
```
marmarta, line 1378: I think this should be based on vm.state, to cut down on direct Admin API calls

qubesmanager/tests/test_qube_manager.py
```diff
@@ -1604,3 +1604,55 @@ def test_704_check_later(mock_timer, mock_question):
 
     assert mock_question.call_count == 0
     assert mock_timer.call_count == 1
+
+
+@pytest.mark.asyncio(loop_scope="module")
+async def test_705_prohibit_start_vms(qubes_manager):
```
marmarta, line 1610: This tests misses the actual "test that something happened" part. In particular, after selecting a VM you are not checking if, for example, the action menu options are appropriate.

marmarta (changes requested):

icons/ban.svg
alimirjamali: Exactly. I fully remembered the color issue. But I did not amend the PR while the OpenQA & review was going on. I will do it shortly.

qubesmanager/qube_manager.py
alimirjamali: Well. This is program internals and not exposed to end-user. We could be less worried about end-user interpretation. I will check it to blocked. But we should remember that the block is only from starting (and not change of properties, features, ...).

p.s.: I will keep the svg name as `ban.svg`.

qubesmanager/qube_manager.py
```diff
@@ -1352,6 +1372,18 @@ def table_selection_changed(self):
             if not vm.updateable and vm.klass != 'AdminVM':
                 self.action_updatevm.setEnabled(False)
 
+            if vm.state['power'] == 'Halted'and \
+                    vm.vm.features.get('prohibit-start', False):
```
marmarta, line 1376:
```suggestion
            if vm.state['outdated'] == 'blocked':
```
Don't use direct API calls here, that's what the whole VmInfo object is for.

qubesmanager/tests/test_qube_manager.py
```diff
@@ -1604,3 +1604,63 @@ def test_704_check_later(mock_timer, mock_question):
...
+                  [('name', 'prohibit-start'),
+                   ('newvalue', 'Do not update')]))
+
+    with contextlib.suppress(asyncio.TimeoutError):
+        await asyncio.wait_for(qubes_manager.dispatcher.listen_for_events(), 1)
+
+    qubes_manager.qubes_app.domains['test-old'].features[ \
+        'prohibit-start'] = \
+        'Compromised by Gremlins!'
+    qubes_manager.qubes_app.domains['test-standalone'].features[ \
```
marmarta, line 1653: That and the line above shouldn't do anything here (also you are setting them a couple of lines above, too).

qubesmanager/qube_manager.py
alimirjamali: Done

qubesmanager/tests/test_qube_manager.py
```diff
@@ -1604,3 +1604,57 @@ def test_704_check_later(mock_timer, mock_question):
 
     assert mock_question.call_count == 0
     assert mock_timer.call_count == 1
+
+
+@pytest.mark.asyncio(loop_scope="module")
+async def test_705_prohibit_start_vms(qubes_manager):
```
marmarta, line 1610: One more test improvement, to make sure it tests the reaction correctly:
- check at the start of the test, before setting of all the features, that the relevant actions are in the correct state, in particular, for those two not-running qubes, action_resumevm should be enabled before you switch the feature.

marmarta: sorry for so many iterations, I keep coming up with ideas how to improve this

alimirjamali: This is also done. Local test is OK. I force pushed right now.

marmarek:
unit test fails

alimirjamali:
> unit test fails

The unittest was correctly detecting that something was wrong. The logic in the Qube Manager was incomplete and only updated `prohibit-start` status for TemplateVMs and StandaloneVMs. This took some time.

qubesmanager/qube_manager.py
```diff
@@ -239,6 +243,12 @@ def update_power_state(self):
 
         self.state['outdated'] = ""
         try:
+            if self.vm.klass != "AdminVM" and manager_utils.get_feature(
+                    self.vm, 'prohibit-start', False):
+                # Special case where being outdated, eol & skipped is irrelevant
```
marmarta, line 248: But a qube with start prohibited might receive updates (a template, for example), I don't think we should ignore that?

marmarek: There is a separate feature to exclude VM from updates. IMO notification about updates should still be visible if not explicitly excluded from updates, because somebody may not realize the implication of prohibit-start thing. Or simply forget about having it enabled (temporarily) and never realize they use outdated template now.

alimirjamali: Ok. Then we have three choices.

1. Use the `power` state column for this. And if the qube is not `Running, Transient, Halting, Paused, Suspended, Dying, Crashed` but `Halted` and have `prohibit-feature` set, then show the ban icon there.
2. We add yet another sub-column to state column. Considering that this column header title is wider than two set of sub-columns (`power` and `outdated` sub-columns), it will not be bad. This approach is nice since we could easily add the `prohibit-start` `rationale` to the mentioned ban icon.
3. Something else

I await your decision and advise on this.

marmarek: IMO the power column makes sense, since prohibit-start VM shouldn't be running (unless started before setting that feature - at which point prohibit-start isn't that relevant yet).

alimirjamali: I amended the PR and moved the ban icon for `prohibit-start` qubes to the power status column. Here is a screenshot of a qube which is both pending update and is start prohibited:

both_prohibit_and_outdated

marmarta (approved):

PR #409, opened 2025-02-19
Unify icon sizes for qube manager icons

marmarta:
Plus a few icon fixes and unified color for monochrome icons (#404040).


marmarta:
@alimirjamali the icon fix we talked about - one note, monochrome icons are not black, they are `#404040`, because it makes them not invisible on dark mode. Not a perfect solution, but before implementing https://github.com/QubesOS/qubes-issues/issues/9784 it should suffice (also I think this color looks quite well).

alimirjamali:
`template-switcher.svg` needs some cleanup as well. If you zoom to drawing (Inkscape Hotkey = 4), you will see the unused curves outside of work area. Same for `checkmark-with-plus.svg`, `some-checked.svg`, `templatevm.svg`

`skipped.svg` document units, width and height are defined in millimeters instead of pixels. And it is not square.
`templatevm.svg` document size is 48x48 instead of of 24x24. Same for `template-switcher.svg`


> (also I think this color looks quite well).

Good.

alimirjamali (approved):
Looks very well on light theme (the icons and manager). Have not checked the qube manager with these new icons on dark theme, but the icons themselves look fine on another system (Vanilla Fedora) with dark theme.

PR #412, opened 2025-03-07
Settings applications tab

alimirjamali:
This PR contains 3 individual commits:

1. Fix for Drag-and-Drop in Qubes VM Settings GUI -> Applications tab
2. PyQt bindings for QImage tinting
3. Icons for Qubes VM Settings GUI -> Applications tab

fixes: https://github.com/QubesOS/qubes-issues/issues/9829

### DIFF
--- a/ci/requirements.txt
+++ b/ci/requirements.txt
@@ -9,3 +9,5 @@ pylint
 sphinx
 PyYAML
 qasync
+Pillow
+numpy
--- a/qubesmanager/appmenu_select.py
+++ b/qubesmanager/appmenu_select.py
@@ -33,7 +33,7 @@ class AppListWidgetItem(QtWidgets.QListWidgetItem):
         else:
             tooltip += "\n" + additional_description
         self.setToolTip(tooltip)
-        self.ident = ident
+        self.setWhatsThis(ident)
         # Using identity as tooltip which also enables drag-and-drop
         self.setWhatsThis(ident)
 
@@ -69,7 +69,7 @@ class AppmenuSelectManager:
             self.whitelisted = []
 
         currently_selected = [
-            self.app_list.selected_list.item(i).ident
+            self.app_list.selected_list.item(i).whatsThis()
             for i in range(self.app_list.selected_list.count())]
 
         whitelist = set(self.whitelisted + currently_selected)
@@ -97,9 +97,9 @@ class AppmenuSelectManager:
             available_appmenus = []
 
         for app in available_appmenus:
-            if app.ident in whitelist:
+            if app.whatsThis() in whitelist:
                 self.app_list.selected_list.addItem(app)
-                whitelist.remove(app.ident)
+                whitelist.remove(app.whatsThis())
             else:
                 self.app_list.available_list.addItem(app)
 
@@ -113,7 +113,7 @@ class AppmenuSelectManager:
         self.app_list.selected_list.sortItems()
 
     def save_appmenu_select_changes(self):
-        new_whitelisted = [self.app_list.selected_list.item(i).ident
+        new_whitelisted = [self.app_list.selected_list.item(i).whatsThis()
                            for i in range(self.app_list.selected_list.count())]
 
         if set(new_whitelisted) == set(self.whitelisted):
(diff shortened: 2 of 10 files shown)


marmarek:
This likely needs --enablerepo flag like here: https://github.com/QubesOS/qubes-core-admin-client/commit/c3759adc91a6a92f3e0927a5ce3459f93900c641

alimirjamali:
> This likely needs --enablerepo flag like here: [QubesOS/qubes-core-admin-client@c3759ad](https://github.com/QubesOS/qubes-core-admin-client/commit/c3759adc91a6a92f3e0927a5ce3459f93900c641)

Did not work

alimirjamali:
OK. Unittests & Pylint are finally green (after hamnering my way to installing qubesimgconverter). The OpenQA installedTest fails most probably because of the icons.

marmarek:
> system_tests_gui_interactive@hw1
> 
>     * simple_gui_apps: [unnamed test](https://openqa.qubes-os.org/tests/131583#step/simple_gui_apps/19) (unknown)
> 
>     * simple_gui_apps: [Failed](https://openqa.qubes-os.org/tests/131583#step/simple_gui_apps/20) (test died)
>       `# Test died: no candidate needle with tag(s) 'vm-settings-app-xterm...`

Yes, all of those are due to different appearance of the applications tab. If Marta says she likes it, I'll adjust base screenshots ("needles"), unfortunately there is no easy way to do that for external contributors.

ci/requirements.txt
```diff
@@ -9,3 +9,5 @@ pylint
 sphinx
 PyYAML
 qasync
+Pillow                                                                                                                              
+numpy
```
marmarek, line 13: A lot of spaces here...

qubesmanager/appmenu_select.py
```diff
@@ -81,25 +93,68 @@ def fill_apps_list(self, template=None):
...
+            main_template = self.vm.template.template.name
+
+        template_icons_path = path.join(
+            path.expanduser("~"),
+            ".local",
+            "share",
+            "qubes-appmenus",
+            f"{main_template}",
+            "apps.tempicons",
+        )
```
marmarek, line 126: Wouldn't it be better to extend `qvm-appmenus` to print the path too? Maybe add some "field" like `OriginalIconPath`?
Especially, the current implementation doesn't handle applications installed in the AppVM itself (you can put applications into `~/.local/share/applications` and they will be available same as those in `/usr/share/applications` in the template).

qubesmanager/appmenu_select.py
```diff
@@ -81,25 +93,68 @@ def fill_apps_list(self, template=None):
...
+                if path.exists(icon_path):
+                    icon = QtGui.QIcon(icon_path)
+                    qpixmap = icon.pixmap(QtCore.QSize(512, 512))
+                    qimage = QtGui.QImage(qpixmap)
+                    qimage = tint_qimage(qimage, self.vm.label.color)
+                    qpixmap = QtGui.QPixmap(qimage)
+                    icon = QtGui.QIcon(qpixmap)
+                else:
+                    # for `qubes-start.desktop` & missing icons
+                    icon = QtGui.QIcon.fromTheme(self.vm.icon)
```
marmarek, line 147: While this fallback case should remain anyway, I think qubes-start.desktop should also have an icon (probably a work for separate PR).

marmarek (changes requested):

ci/requirements.txt
alimirjamali: Fixed.

qubesmanager/appmenu_select.py
alimirjamali: P.S. One more consideration. The GUIVM icons in `~/.local/share/qubes-appmenus/VMNAME/apps.icons` provided via qvm-appmenus are already tinted (applicable for Qube specific Flatpak, Snap, ...). And we have to consider this as well if we ever want to patch `qvm-appmenus` to provide real original icon.

marmarek: > The GUIVM icons in `~/.local/share/qubes-appmenus/VMNAME/apps.icons`

Yes, but there is also `apps.tempicons` dir, similar to icons from the template.

marmarek (approved):

PR #414, opened 2025-03-12
Add `prohibit-start` to Qube Settings GUI.

alimirjamali:


marmarek:
This behaves funny regarding empty value. First, when enabling the checkbox but not entering anything, the error message is "Error while saving changes: Feature not set for domain sys-whonix: prohibit-start". It isn't exactly clear that it's about empty value.
But it's also interesting the other way around: when it's set with some rationale, and then you remove the rationale, the checkbox automatically uncheck and the entry box disappears. Pretty annoying if you try to just change the rationale...

alimirjamali:
> This behaves funny regarding empty value. First, when enabling the checkbox but not entering anything, the error message is "Error while saving changes: Feature not set for domain sys-whonix: prohibit-start". It isn't exactly clear that it's about empty value. But it's also interesting the other way around: when it's set with some rationale, and then you remove the rationale, the checkbox automatically uncheck and the entry box disappears. Pretty annoying if you try to just change the rationale...

Yes. I believe that I have to make it better. I believe once the rationale TextEntry box is shown, it should remain visible regardless of the status of the checkbox. And the logic has to be improved as well.

marmarek:
Hiding it when checkbox gets unchecked is probably okay too (and probably should keep the value until dialog is closed). It the checkbox shouldn't automatically uncheck when rationale is cleared...

marmarta (changes requested):
So, on testing, I think there is a bit of UX roughness here: namely, a user can click prohibit start and forget to provide reason, plus, the reason entry box is a bit awkward. My proposal:
- provide a default reason
-  move the text box next to the checkbox, to avoid moving things around when the checkbox is clicked
- if the text box is empty and visible, it should scream in some way - maybe a red border? - and ok/apply probably shouldn't be clickable

alimirjamali:
> * provide a default reason
> * move the text box next to the checkbox, to avoid moving things around when the checkbox is clicked
> * if the text box is empty and visible, it should scream in some way - maybe a red border? - and ok/apply probably shouldn't be clickable

- A default reason is added.
- The checkbox and textbox were in a QHBoxLayout. I am not sure why I switched to QVBoxLayout. Switched back to VBox.
- Red border was a little bit troublesome. I made the placeholder text red instead. It looks good IMO (see video).

https://github.com/user-attachments/assets/5b99f510-6f79-4712-8a9b-dd74ecb5b7f5

marmarta (approved):
I love it!

marmarta (approved):

PR #417, opened 2025-03-19
Fix `qubes-vm-boot-from-device`, Improve settings

alimirjamali:


marmarek:
Now you need to do something about pylint:
```
qubesmanager/bootfromdevice.py:35:4: R0917: Too many positional arguments (6/5) (too-many-positional-arguments)
```
...

alimirjamali:
> Now you need to do something about pylint:
> 
> ```
> qubesmanager/bootfromdevice.py:35:4: R0917: Too many positional arguments (6/5) (too-many-positional-arguments)
> ```
> 
> ...

Already did it (before seeing your message).

marmarek (approved):

PR #419, opened 2025-03-24
Improve and fix Settings App icons

alimirjamali:
Set a blank icon for missing apps. Scale up small icons

### DIFF
--- a/qubesmanager/appmenu_select.py
+++ b/qubesmanager/appmenu_select.py
@@ -140,6 +140,11 @@ class AppmenuSelectManager:
                     qpixmap = icon.pixmap(QtCore.QSize(512, 512))
                     qimage = QtGui.QImage(qpixmap)
                     qimage = tint_qimage(qimage, self.vm.label.color)
+                    if (
+                        qimage.size().width() < 48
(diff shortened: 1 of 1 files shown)


alimirjamali:
Making pylint and black happier

marmarek:
This sets a qube icon for the "Start qube" entry, but that one has a different icon in the qubes menu (which also isn't colored, as it's one of the special entries at the bottom).
OTOH, in a standard xfce menu, it is indeed the qube icon, so maybe that's okay?

alimirjamali:
> This sets a qube icon for the "Start qube" entry, but that one has a different icon in the qubes menu (which also isn't colored, as it's one of the special entries at the bottom). OTOH, in a standard xfce menu, it is indeed the qube icon, so maybe that's okay?

Yes. I believe this is OK. If we want the ▶️ icon or something similar, the code generating the "Start qube" .desktop file should add it as its icon. But I can amend the PR and add special case if it is necessary. Please advise

marmarek:
Makes sense to simply change it in the desktop file and not have a special case here.

marmarek (approved):

PR #420, opened 2025-04-16
Allow user to select compression filter

alimirjamali:
Provide user a selection of default compression filters in addition to the available recognised filters. Allow user to save the filter.

Requires: https://github.com/QubesOS/qubes-core-admin-client/pull/346

Related: https://github.com/QubesOS/qubes-issues/issues/8211
Related: https://github.com/QubesOS/qubes-issues/issues/8291

### DIFF
--- a/qubesmanager/backup.py
+++ b/qubesmanager/backup.py
@@ -23,6 +23,8 @@
 import signal
 from qubesadmin import exc
 from qubesadmin import utils as admin_utils
+from qubesadmin.backup.restore \
+    import KNOWN_COMPRESSION_FILTERS, OPTIONAL_COMPRESSION_FILTERS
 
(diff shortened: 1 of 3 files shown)


marmarek:
This doesn't work: https://openqa.qubes-os.org/tests/137193#step/qubesmanager_backuprestore/9
There is no default selected, and the list appears to be empty.

alimirjamali:
> This doesn't work: https://openqa.qubes-os.org/tests/137193#step/qubesmanager_backuprestore/9 There is no default selected, and the list appears to be empty.

I am trying to figure out what is wrong here. It appears that the UI file is applied but the code is not.

marmarek (approved):

PR #422, opened 2025-04-25
anon gateway warning box on main qubesmanager page

Guiiix:
Solves:

> Please re-open.
> 
> [A user reported](https://forums.whonix.org/t/re-stop-users-from-changing-their-anon-whonix-net-qube-to-sys-firewall-to-avoid-ip-leaks/17255/7):
> 
> I remember looking at this and patching my Qubes Manager temporarily and it failed to account for switching NetVM in the basic view. The pop up only shows in the settings tab, meaning this is only half solved
> 
> This should be verified and reported before 4.3 is out, I have not checked the code again 

 _Originally posted by @adrelanos in [#8551](https://github.com/QubesOS/qubes-issues/issues/8551#issuecomment-2797290629)_


qubesmanager/qube_manager.py
```diff
@@ -969,27 +998,20 @@ def change_network(self, netvm_name):
...
-            if check_power and netvm and not netvm.is_running():
-                reply = QMessageBox.question(
-                    self, self.tr("Qube Start Confirmation"),
-                    self.tr("<br>Can not change netvm to a halted Qube.<br>"
-                        "Do you want to start the Qube <b>'{0}'</b>?").format(
-                        netvm_name),
-                    QMessageBox.StandardButton.Yes |
-                    QMessageBox.StandardButton.Cancel)
+        # pylint: disable=possibly-used-before-assignment
+        if netvm_name and check_power and netvm and not netvm.is_running():
```
marmarek, line 1002: This can be simplified to avoid checking both `netvm_name` and `netvm`, you just need to initialize `netvm = None` at the start of the function. This will also avoid the pylint complain (you may also need to initialize `check_power` too).

marmarek (changes requested):

marmarek (approved):

PR #424, opened 2025-05-03
Show CPU/memory usage in Qube manager again

### LINKED ISSUE #7817: Show CPU/memory usage in Qube manager again
[Back in 2018](https://github.com/QubesOS/qubes-manager/pull/52) the CPU and memory usage columns were removed from the Qube manager.
Since then it annoys me I only see these values in the widget. I see it as a regression as one could select which columns to display in the "View" tab of Qube manager. Please reintroduce these.

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/7817

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -244,10 +244,18 @@ class VmInfo():
         self.state = {'power': "", 'outdated': ""}
         self.updateable = getattr(vm, 'updateable', False)
         self.update(True)
+        self.CPU_usage = None
+        self.RAM_usage = None
+
+    def update_resource_usage(self, RAM_usage, CPU_usage):
+        self.RAM_usage = RAM_usage
+        self.CPU_usage = CPU_usage
 
     def update_power_state(self):
         try:
             self.state['power'] = self.vm.get_power_state()
+            if self.state["power"] in ["Halted", "Paused", "Suspended"]:
+                self.update_resource_usage(None, None)
             if self.state['power'] == "Halted" and \
                     self.vm.klass != "AdminVM" and \
                     manager_utils.get_feature(
@@ -258,6 +266,7 @@ class VmInfo():
                 self.state['power'] = 'Blocked'
         except exc.QubesDaemonAccessError:
             self.state['power'] = ""
+
 
         self.state['outdated'] = ""
         try:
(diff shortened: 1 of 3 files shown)


marmarta (changes requested):
I played around with this PR for a bit, and while I generally love it, I have comments:
- sorting: I think that sorting on usage is kinda problematic right now; it keeps dom0 as top item, which is a bit counterintuitive but consistent with the rest of the code, but also on ascending sorting puts all the non-running VMs on top and the running ones on bottom, which is definitely unintuitive, and on descending sorting puts dom0 at the very end, even though it is definitely running. I think any sorting on usage should put VMs that are consuming resources on top, disregarding special dom0 status too.
- I get different numbers on CPU usage in qube manager and in domains widget, which is not great; there seems to be different rounding, because the error is off-by-one sometimes, but I think they need to be consistent

Sahilll10:
Hi @alimirjamali @marmarta 
I am working on the System Health Monitor project for GSoC 2026, and in order to do some groundwork before the Pre-GSOC period, I wanted to ask if I can pick up the sorting fix that marmarta flagged.

After thoroughly going through the files and code, the sorting for CPU and MEM columns currently returns int(vm.CPU_usage) directly.

The problem is:
Non-running VMs return None for CPU_usage so they end up sorted incorrectly above running VMs in ascending order
dom0 has no special handling so it falls to the bottom in descending order

The fix I am thinking of — 1) for CPU and MEM sort values, return -1 for non-running VMs so they always sort below running ones, and keep dom0 pinned at the top the same way it is handled in the existing UserRole + 1 sort logic.
2) For the rounding difference between qube_manager and domains widget — domains widget uses int(cpu_usage) directly from vm-stats. 
The PR rounds the same way so I need to check where the off-by-one comes from.

Would it be okay if I open a PR with this fix building on your work? I will be testing it using MockQubesComplete.

ben-grande:
> * I get different numbers on CPU usage in qube manager and in domains widget, which is not great; there seems to be different rounding, because the error is off-by-one sometimes, but I think they need to be consistent

I think the issue is that data is collected at different moments. Each subscriber to `admin.vm.Stats` respect `app.stats_interval` (see `qubes.admin.api`), so if they query at different moments, the time elapsed compared to the amount of CPU time spent by a qube will be different.

One idea to get consistent results is to cache the results, and return the same information to all clients.

PR #427, opened 2025-06-07
Show (Manager App) Icons for Wayland/KDE

alimirjamali:
Without this, generic Wayland icon will be shown


alimirjamali:
@marmarek The name of `Qubes` app object was not consistent with others in `settings`. I made a mistake there. The openQA tests need to be restarted.

alimirjamali:
PipelineRetryFailed

PR #430, opened 2025-09-27
Add Application search to qube settings

alimirjamali:
resolves: https://github.com/QubesOS/qubes-issues/issues/10276

### DIFF
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -261,6 +261,8 @@ class VMSettingsWindow(ui_settingsdlg.Ui_SettingsDialog, QtWidgets.QDialog):
             self.app_list_manager = AppmenuSelectManager(self.vm, self.app_list)
             self.refresh_apps_button.clicked.connect(self.refresh_apps_button_pressed)
 
+            self.app_search.textChanged.connect(self.filter_apps)
+
             # Enable Drag & Drop between between two panels
             # ToDo: Disable D&D between multiple instances of qubes-vm-settings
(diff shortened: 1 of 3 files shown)


qubesmanager/settings.py
```diff
@@ -1684,6 +1686,22 @@ def template_apps_change(self):
                 self.app_list_manager.has_missing
             )
 
+    def filter_apps(self):
```
marmarta, line 1689: the problem with this approach is that if someone searches for "xfce  terminal" or "terminal xfce" they won't find either. I think we should find things matching every word in the search query?

qubesmanager/tests/test_vm_settings.py
```diff
@@ -1922,6 +1922,21 @@ def test_506_application_remove_all(settings_fixture):
         assert expected_call not in settings_window.qubesapp.actual_calls
 
 
+@check_errors
+@pytest.mark.parametrize("settings_fixture", ["test-red"], indirect=True)
+def test_507_filter_apps(settings_fixture):
```
marmarta, line 1927: looks good but I'd like also tests for two words with two spaces between and two words in reverse order (as mentioned above)

marmarta (changes requested):

qubesmanager/tests/test_vm_settings.py
alimirjamali: This is also done. I also added leading and trailing spaces

PR #431, opened 2025-09-28
Add notes counter

alimirjamali:


alimirjamali:
Someone asked for this during Qubes OS Summit 2025.

It is not that much necessary IMHO. As 4KB is roughly one A4 page of text. But It won't hurt.

marmarek (approved):

PR #440, opened 2025-12-09
Improve reactions to missing PCI devices

### LINKED ISSUE #10409: Qube Manager crashes when attempting to view settings for a qube containing PCI devices that no longer exist
[How to file a helpful issue](https://www.qubes-os.org/doc/issue-tracking/)

### Qubes OS release

R4.3

### Brief summary

Yesterday, I ended up moving my testing Qubes OS installation from a fairly recent Intel-based system (13th gen CPU) to a somewhat older one (11th gen CPU). The USB controller in this system is different than the one in the previous system, and so sys-usb would not start since it couldn't pass through a non-existent PCI device. To resolve this, I assumed I'd be able to open Qube Manager, select sys-usb, click Settings, then go to the Devices tab and remove the non-existent device. Unfortunately, upon clicking the "Settings" button, Qube Manager crashed.

### Steps to reproduce

1. Install Qubes OS R4.3 on a machine with a particular generation of Intel CPU (I assume AMD CPUs will show the same problems).
2. Boot the installed system once, verify sys-usb and sys-net start, then shut the system down properly.
3. Move the drive from the system where you installed it to some other system with a different generation of Intel CPU.
4. Boot the installed system. sys-usb should not start.
5. Attempt to start sys-usb. You should see an error about a PCI device not being available.
6. In dom0, open xfce4-terminal, and run `qubes-qube-manager`.
7. Click on sys-usb.
8. Click the Settings button.

### Expected behavior

VM settings window appears.

### Actual behavior

Qube Manager crashes, and a Python exception is printed in the terminal.

### Additional information

I was able to recover from this state by running `sudo qvm-pci unassign sys-usb`. After this, the settings window would open normally for sys-usb.

marmarta:
Fix tests so that now they will fail if there is a bad reaction to an unknown PCI device. Clean-up handling unknown PCI devices by settigs.

requires https://github.com/QubesOS/qubes-core-admin-client/pull/400
 fixes QubesOS/qubes-issues#10409

### DIFF
--- a/qubesmanager/settings.py
+++ b/qubesmanager/settings.py
@@ -1551,15 +1551,13 @@ class VMSettingsWindow(ui_settingsdlg.Ui_SettingsDialog, QtWidgets.QDialog):
 
         # pylint: disable=too-few-public-methods
         class DevListWidgetItem(QtWidgets.QListWidgetItem):
-            def __init__(self, dev, unknown=False, parent=None):
+            def __init__(self, dev, parent=None):
                 super().__init__(parent)
                 name = (
                     dev.port_id.replace("_", ":").replace("-", " -> ")
                     + " "
                     + dev.description
                 )
-                if unknown:
-                    name += " (unknown)"
                 self.setText(name)
                 self.dev = dev
                 intfs = list({i.category for i in dev.interfaces})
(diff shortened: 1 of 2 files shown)


marmarek:
PipelineRetry

marmarek:
PipelineRetry

marmarek (approved):

PR #443, opened 2026-01-13
Hide internal qubes from boot device selection

ben-grande:
For: https://github.com/QubesOS/qubes-issues/issues/1512

### DIFF
--- a/qubesmanager/bootfromdevice.py
+++ b/qubesmanager/bootfromdevice.py
@@ -108,12 +108,7 @@ class VMBootFromDeviceWindow(ui_bootfromdevice.Ui_BootDialog,
         utils.initialize_widget_with_vms(
             widget=self.fileVM,
             qubes_app=self.qubesapp,
-            filter_function=(lambda vm:
-                             vm != self.vm
-                             and vm.klass != "TemplateVM"
-                             and not utils.get_feature(vm, "internal", False)
-                            ),
-            allow_internal=True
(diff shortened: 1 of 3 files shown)


qubesmanager/bootfromdevice.py
```diff
@@ -108,7 +108,14 @@ def __init_buttons__(self):
...
-            filter_function=(lambda x: x != self.vm),
+            filter_function=(lambda vm:
+                             vm.klass != "TemplateVM"
+                             and not utils.get_feature(vm, "internal", False)
+                             and not utils.get_feature(vm, "servicevm", False)
+                             and not utils.get_feature(vm, "service.audiovm", False)
+                             and not utils.get_feature(vm, "service.guivm", False)
+                             and not getattr(vm, "template_for_dispvms", False)
+                            ),
             allow_internal=True
```
marmarek: The commit message talks about hiding internal. There is already parameter for that just here, instead of the whole filter function. And IMO filtering-out all those (other) qubes isn't a good idea - for example excluding servicevm excludes sys-usb, which is a very valid choice (for example to boot from a USB stick plugged in).

marmarek (changes requested):

marmarek:
test_10_do_backup fails now, but later on tests hang, so there is no full details of the failure...

marmarek (changes requested):

marmarek (approved):

PR #447, opened 2026-03-27
firewall: Add comment field to firewall rules

### LINKED ISSUE #952: Firewall rule "comment" field
**Reported by axon on 16 Feb 2015 06:11 UTC**
It would be useful to have a plain text "comment" or "note" field when creating individual firewall rules (similarly to #899). Here are some examples of useful information a user might decide to put there:
- The domain name corresponding (at the time) to a bare IP address or CIDR block. (Useful to save on whois searches later. Also useful if the rule ever stops working, so the user knows what was originally intended.)
- The user's reason for adding that firewall rule. (Useful when wondering, later, why that rule exist.)
- Miscellaneous notes to self.

(Note: I'm restricted from changing the priority of this ticket from "major" to "minor.")

Migrated-From: https://wiki.qubes-os.org/ticket/952

nihalxkumar:
Allows users to add optional plain-text comments/notes to individual firewall rules for documentation purposes. The comment is metadata only and does not affect the actual firewall behavior.

I have tested the GUI on Arch Linux:

screenshot of qubes manager GUI showcasing the comment label

closes: https://github.com/QubesOS/qubes-issues/issues/952

### DIFF
--- a/ui/newfwruledlg.ui
+++ b/ui/newfwruledlg.ui
@@ -116,10 +116,10 @@
       <item row="3" column="1" colspan="3">
       <widget class="QLineEdit" name="commentLineEdit">
        <property name="toolTip">
-        <string>Optional note for your own reference.</string>
+        <string>Comments are for user reference only</string>
        </property>
        <property name="placeholderText">
-        <string>Optional note (e.g. reason for this rule)</string>
+        <string>Comment (optional)</string>
        </property>
        <property name="maxLength">
         <number>512</number>
--- a/ui/newfwruledlg.ui
+++ b/ui/newfwruledlg.ui
@@ -10,7 +10,7 @@
     <x>0</x>
     <y>0</y>
     <width>381</width>
-    <height>193</height>
+    <height>225</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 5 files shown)


ui/newfwruledlg.ui
```diff
@@ -105,8 +105,28 @@
...
+      <widget class="QLabel" name="label_5">
+       <property name="text">
+        <string>Comment</string>
+       </property>
+      </widget>
+      </item>
+      <item row="3" column="1" colspan="3">
+      <widget class="QLineEdit" name="commentLineEdit">
+       <property name="toolTip">
+        <string>Optional note for your own reference.</string>
```
marmarta:
```suggestion
        <string>Comments are for user reference only</string>
```

ui/newfwruledlg.ui
```diff
@@ -105,8 +105,28 @@
...
+       </property>
+       <property name="placeholderText">
+        <string>Optional note (e.g. reason for this rule)</string>
```
marmarta:
```suggestion
        <string>Comment (optional)</string>
```

marmarta (changes requested):
I like this. Minor wording adjustments to be a bit simpler.

ui/newfwruledlg.ui
nihalxkumar: Yes, this would be better.

ui/newfwruledlg.ui
nihalxkumar: Thanks

PR #449, opened 2026-04-27
Make Template Manager text selectable for copying

coyotebush:
In particular, the ability to copy the installed template version would be useful.

I assume this change compiles to the following line which I've manually inserted into the compiled `ui_qvmtemplate.py`:
```py
self.template_info.setTextInteractionFlags(QtCore.Qt.TextInteractionFlag.LinksAccessibleByMouse|QtCore.Qt.TextInteractionFlag.TextSelectableByMouse)
```

### DIFF
--- a/ui/qvmtemplate.ui
+++ b/ui/qvmtemplate.ui
@@ -62,6 +62,9 @@
              <property name="text">
               <string>lorem ipsum</string>
              </property>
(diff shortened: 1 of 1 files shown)


marmarta (approved):
that's a good idea (also yeah, that's the correct change)

PR #450, opened 2026-05-12
Don't block main thread on qube hanging shutdown

### LINKED ISSUE #10074: DoS: Qube Manager locks up when qube does not respond to its commands
### Qubes OS release

Qubes OS 4.2, Qubes OS 4.3

### Brief summary

Frequently, I will have to stop a stuck qube. I had thought this was only the case when qubes are frozen, however, this time my Personal browser qube was still allowing me to use Firefox and Thunar, without issue.

The only trouble I was having was a flash drive refusing to attach to this qube, after having connected several on/off earlier this boot.

When trying to shutdown this qube from Qube Manager, I was declined:

Image

Whenever Qube Manager gets such an error, the entirety of Qube Manager does not respond to any input

This includes even trying to spawn a second Qube Manager, which will instead queue along with any other actions, until the process times out. 

### Steps to reproduce

1. Get any qube in a stuck state
2. Attempt to shutdown or restart this qube from Qube Manager
3. Observe as Qube Manager is locked up for at least 60 seconds

### Expected behavior

Qube Manager should only lock up an existing process as pending, maybe giving some visual feedback in-line or for those buttons

This should be overridable at any time with Kill, before the initially requested command resolves

### Actual behavior

Qube Manager is frustratingly locked up for a long time

### Additional information

_No response_

ben-grande:
The merge request to core-admin already avoids the main problem, of a hanging             shutdown blocking qubesd. This change avoids a hanging shutdown blocking the              interaction with the qube manager

For: https://github.com/QubesOS/qubes-issues/issues/10648
Fixes: https://github.com/QubesOS/qubes-issues/issues/10074
Requires: https://github.com/QubesOS/qubes-core-admin/pull/807
Requires: https://github.com/QubesOS/qubes-core-admin-client/pull/469

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -57,10 +57,10 @@ const-rgx=(([A-Za-z_][A-Za-z0-9_]*)|(__.*__))$
 class-rgx=([A-Z_][a-zA-Z0-9]+|TC_\d\d_[a-zA-Z0-9_]+)$
 
 # Regular expression which should only match correct function names
-function-rgx=[a-z_][a-z0-9_]{2,30}$
+function-rgx=[a-z_][a-z0-9_]{2,40}$
 
 # Regular expression which should only match correct method names
-method-rgx=[a-z_][a-z0-9_]{2,30}$
+method-rgx=[a-z_][a-z0-9_]{2,40}$
 
 # Regular expression which should only match correct instance attribute names
 attr-rgx=[a-z_][a-z0-9_]{2,30}$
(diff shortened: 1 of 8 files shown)


ben-grande:
Requires bumping qubesadmin in the requirements.

ben-grande:
PipelineRetryFailed

marmarta:
There are some pylint fails and test complaints, otherwise this looks fine

PR #455, opened 2026-07-15
Replace Transient state for Starting and Halting

### LINKED ISSUE #10966: Improve handling of `Transient` power state
PRs:

- https://github.com/QubesOS/qubes-core-admin/pull/819 (some commits)
- https://github.com/QubesOS/qubes-core-admin-client/pull/480
- https://github.com/QubesOS/qubes-desktop-linux-menu/pull/73
- https://github.com/QubesOS/qubes-desktop-linux-manager/pull/321
- https://github.com/QubesOS/qubes-manager/pull/455
- https://github.com/QubesOS/qubes-mgmt-salt-dom0-qvm/pull/40

---

The `Transient` power state is not really a domain state, at least not something know by Xen or Libvirt. It is something that Qubes uses for a running domain with Qrexec available but not running. This creates all kinds of weird issues:

- Developers don't think of it as a real power state, as you can see from the linked issues:
  - Used as a [substitute for `Halting`](https://github.com/QubesOS/qubes-desktop-linux-manager/blob/7ab4e2e043408c494e78e4aad7dc10ae2dd66d95/qui/tray/domains.py#L49), with [`Halting` not really working yet](https://github.com/QubesOS/qubes-core-admin/pull/819), but still, this is not the same as Qrexec not connected, because `domain-pre-shutdown` can fail and the VM should never have been `Transient`, but `Halting`.
  - Used as a [substitute for failing to get the real power state](https://github.com/QubesOS/qubes-desktop-linux-manager/blob/7ab4e2e043408c494e78e4aad7dc10ae2dd66d95/qui/tray/domains.py#L1183).
  - Not considered when [`unpausing` a domain](https://github.com/QubesOS/qubes-core-admin-client/blob/b8a948d9a6e511282fac909c0fac051718ab0746/qubesadmin/app.py#L816), which defaults to `Running`, which is not always correct
  - Not considered in a [recent PR to the server to cache domain state](https://github.com/QubesOS/qubes-core-admin/pull/819), with `unpause` and `resume` always leading to `Running`
- It is a state only know by the server, as it needs to check for Qrexec file on AdminVM. As this is not really a power state, watching libvirt events or querying via the API is not enough, something like `inotify` could be used. But still, this leads to other issues:
  - As the state is only known by the server, when inotify runs and there is an update, we need to notify the clients of the new state, a new event for notifying that a `Running` domain became `Transient` as Qrexec is not connected.
  - The server has to broadcast the state as an event argument on certain state changes, such as `unpause` and `resume`, to pass down to the clients the state if it is `Transient`.

Because of all these problems, we are considering deprecating the `Transient` state. I understand some client tools use this state to decorate the qube, the qui-domain use it with a red color and loading gear icon to indicate that the qube is not in a "ready" state. For `Halting` I am fixing the PR mentioned above. For a `Starting` one, we might add it.

Another solution is:

- Make all clients use the same method on core-admin-client to get a consistent power state
- Monitor qrexec connection and fire event on change:
  - `inotify` the Qrexec socket on the server
  - `domain-feature-(set|delete):qrexec` can also impact
  - But don't fire the even in all states, or pass the state as an argument, because a qube that wants to shutdown, should not go or doesn't need to go from `Running -> Halting -> Transient -> Halted`, which might be similar to what was intended back in the day as per this [graph](https://github.com/QubesOS/qubes-core-admin/blob/0bea4d6ae7df26ceb0f8d91cd8b5e1aff7e9849c/qubes/vm/qubesvm.py#L2438), but that is far from the case of what happens today, and informing `Transient` doesn't really help here.
- Treat `Transient` as a property rather than a power state. On property change, fire appropriate event. Which in some sense feels correct, as it is not really a state, and also would help avoid the client from guessing or the server to fire a new event or pass it down via event argument.
- Introduce a `Starting` state, which similar to the `Halting` state, is a Qube state, not known by libvirt or Xen, but contratry to `Transient`, it can be fully dealt with using `domain-pre-star`->`Starting`, `domain-start-failed`->`Halted`, `domain-start`->`Running`. Technically, `Halting` could be a libvirt state, but that even is not being fired by libxl. Then, clients would know how to display this non-ready state instead of a hard switch from `Halted` to `Running`.

ben-grande:
For: https://github.com/QubesOS/qubes-issues/issues/10966
Requires: https://github.com/QubesOS/qubes-core-admin-client/pull/480

### DIFF
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -100,6 +100,7 @@ class StateIconDelegate(QStyledItemDelegate):
                 "Paused" : QIcon(":/paused"),
                 "Suspended" : QIcon(":/paused"),
                 "Transient" : QIcon(":/transient"),
+                "Starting" : QIcon(":/transient"),
                 "Halting" : QIcon(":/transient"),
                 "Dying" : QIcon(":/transient"),
                 "Halted" : QIcon(":/blank"),
@@ -549,8 +550,8 @@ class QubesTableModel(QAbstractTableModel):
                 state = vm.state.get('power', '')
                 try:
                     ordered_state = str(
-                        ["Running", "Transient", "Halting", "Paused",
-                         "Suspended", "Dying", "Crashed",
+                        ["Running", "Transient", "Starting", "Paused",
+                         "Suspended", "Dying", "Halting", "Crashed",
                          "Halted", "NA"].index(state))
                 except ValueError:
                     ordered_state = state
@@ -1395,7 +1396,7 @@ class VmManagerWindow(ui_qubemanager.Ui_VmManagerWindow, QMainWindow):
             #  TODO: add boot from device to menu and add windows tools there
             # Update available actions:
             if vm.state['power'] in \
-                    ['Running', 'Transient', 'Halting', 'Dying']:
+                    ['Running', 'Transient', 'Starting', 'Halting', 'Dying']:
                 self.action_resumevm.setEnabled(False)
                 self.action_removevm.setEnabled(False)
                 self.template_menu.setEnabled(False)
--- a/qubesmanager/qube_manager.py
+++ b/qubesmanager/qube_manager.py
@@ -1406,7 +1406,7 @@ class VmManagerWindow(ui_qubemanager.Ui_VmManagerWindow, QMainWindow):
                 self.action_restartvm.setEnabled(False)
                 self.action_open_console.setEnabled(False)
                 self.template_menu.setEnabled(False)
-            elif vm.state['power'] == 'Suspend':
+            elif vm.state['power'] == 'Suspended':
                 self.action_removevm.setEnabled(False)
                 self.action_pausevm.setEnabled(False)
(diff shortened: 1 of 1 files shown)


ben-grande:
Without this PR, the application crashes when it sees a power state it is not aware of, such as `Starting`. This PR fixes it, but for next time, each client application should handle unknown states gracefully.

```python3
Jul 16 17:06:21 dom0 python3[186072]: Traceback (most recent call last):
                                        File "/usr/lib/python3.13/site-packages/qubesmanager/qube_manager.py", line 173, in paint
                                          self.stateIcons[index.data()['power']].pixmap(iconSize))
                                          ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
                                      KeyError: 'Starting'
Jul 16 17:06:21 dom0 systemd-coredump[187813]: [] Process 186072 (qubes-qube-mana) of user 1000 dumped core.
```

Is a transition period necessary? Where the server sends `Transient` in place of `Starting`?

PR #457, opened 2026-07-27
Get attribute from VM object instead of string

### LINKED ISSUE #11015: Booting from ISO broken in qube settings menu
### Qubes OS release

Qubes OS 4.3

### Brief summary

Trying to boot qube from ISO from qube settings menu results in error.

### Steps to reproduce

1. Open qube settings menu
2. Click "Boot qube from DISC or block device"
3. Get the following error:
```
Whoops. A critical error has occurred. This is most likely a bug in Qubes Manager.

AttributeError: 'str' object has no attribute 'klass'
at line 1391
of file /usr/lib/python3.13/site-packages/qubesmanager/settings.py.
```
4. Click "Show Details..."
5. Observe the following:
```
----
line: if vm.klass in ["RemoteVM", "AdminVM"]:
func: __init__
line no.: 46
file: /usr/lib/python3.13/site-packages/qubesmanager/bootfromdevice.py
----
line: boot_dialog = bootfromdevice.VMBootFromDeviceWindow(
func: boot_from_cdrom_button_pressed
line no.: 1391
file: /usr/lib/python3.13/site-packages/qubesmanager/settings.py
```

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional information

All dom0 repos enabled
qubes-manager 4.3.23
Using `qvm-start --cdrom` works

ben-grande:
Fixes: https://github.com/qubesos/qubes-issues/issues/11015

### DIFF
--- a/qubesmanager/bootfromdevice.py
+++ b/qubesmanager/bootfromdevice.py
@@ -27,14 +27,24 @@ from PyQt6 import QtWidgets, QtGui, QtCore  # pylint: disable=import-error
 from qubesadmin import tools
 from qubesadmin import exc
 from qubesadmin import utils as admin_utils
+from qubesadmin.app import QubesBase
+from qubesadmin.vm import QubesVM
 
 # this is needed for icons to actually work
 # pylint: disable=unused-import, no-name-in-module
 from . import resources
 
-class VMBootFromDeviceWindow(ui_bootfromdevice.Ui_BootDialog,
-                             QtWidgets.QDialog):
-    def __init__(self, vm, qapp, qubesapp=None, *, parent=None, new_vm=False):
+
+class VMBootFromDeviceWindow(ui_bootfromdevice.Ui_BootDialog, QtWidgets.QDialog):
+    def __init__(
+        self,
+        vm: QubesVM,
+        qapp: QtWidgets.QApplication,
+        qubesapp: QubesBase,
+        *,
+        parent=None,
+        new_vm: bool = False
+    ):
         super().__init__(parent)
 
         self.vm = vm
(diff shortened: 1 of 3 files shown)


qubesmanager/bootfromdevice.py
```diff
@@ -27,27 +27,33 @@
...
 
 # this is needed for icons to actually work
 # pylint: disable=unused-import, no-name-in-module
 from . import resources
 
 class VMBootFromDeviceWindow(ui_bootfromdevice.Ui_BootDialog,
                              QtWidgets.QDialog):
-    def __init__(self, vm, qapp, qubesapp=None, *, parent=None, new_vm=False):
+    def __init__(
+        self, vm: QubesVM, qapp: QtWidgets, *, parent=None, new_vm: bool = False
```
marmarek: The argument change needs updating all places where it's used. For example, you missed the standalone `qubes-vm-boot-from-disk` command (and BTW there deviating from standardized arguments makes it problematic...).

marmarek: And also, I question type type hints: `QtWidgets` is a module name, I think you meant `QtWidgets.QApplication`.

qubesmanager/bootfromdevice.py
```diff
@@ -27,27 +27,33 @@
...
 
         self.vm = vm
         self.qapp = qapp
-        self.qubesapp = qubesapp
+        self.qubesapp = self.vm.app
         self.cdrom_location = None
         self.new_vm = new_vm
 
-        if vm.klass in ["RemoteVM", "AdminVM"]:
+        if self.vm.klass in ["RemoteVM", "AdminVM"]:
```
marmarek, line 56: Any specific reason for this change (here and below)?

marmarek (changes requested):

qubesmanager/bootfromdevice.py
ben-grande: Mostly stylistic choice, that when using a variable that is a class attribute, I prefer using it rather than the parameter.

ben-grande: Also see on line 61, it using `self.vm` instead of `vm`...

marmarek (approved):

